PR #10, opened 2015-01-21
Fix a typo.

bangerth:
Really nothing of substance, but I noticed it while reading your code :-)


bangerth:
Took only a year and a month :-)

cburstedde:
> Took only a year and a month :-)

Yes (I had merged it into develop way earlier).  ;)

Will try to speed up my response by a factor of two every time.  Please
continue sending pull requests if you have one.

PR #18, opened 2016-09-08
Remove unnecessary NULL checks on arrays

lcw:
This was getting flagged by clang with a warning.


cburstedde:
Thanks. Merged into develop. Pulled into master whenever that one is bumped.

PR #26, opened 2017-05-16
Instruction for Windows build and linking with MS Visual Studio

mike-matveev:
Instructions for Windows build using MSYS2/MinGW64 toolchain.

### DIFF
--- a/INSTALL_WINDOWS
+++ b/INSTALL_WINDOWS
@@ -1,0 +1,84 @@
+Compiling for Windows
+=====================
+
+  It is possible to compile p4est for Windows and build a dll-library for use 
+in an external project within Microsoft Visual Studio toolchain.
+
+Instructions:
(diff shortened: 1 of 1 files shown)


cburstedde:
Mike, this is great. We could add a --enable-windows configure option. By your suggestions, it would
* enable the changes to the source code
* add make targets to create the dll/lib files.
I can add the first part, but for the second I'd rely on your testing. Do you want to give it a shot?

mike-matveev:
Carsten, ok, I will try to do it but cannot promise it is going to be very soon. Also, please note that this effectively kills all the multi-thread features of the library.

cburstedde:
Cool, whenever you get to it -- you could make your windows build crash with --enable-pthread or --enable-openmp. Without these flags it should work (no threading used then).

cburstedde:
Please create pull requests against develop next time.
Thanks and merged.

PR #42, opened 2018-06-01
Allowing user to preallocated p4est_t/p8est_t

jkozdon:
In an application I'm working on it was convenient to allow the user to be in charge of allocating the p4est_t/p8est_t. Thought this might be of use to others.

### DIFF
--- a/doc/author_kozdon.txt
+++ b/doc/author_kozdon.txt
@@ -1,0 +1,1 @@
+I place my contributions to p4est under the CC0 license. Jeremy E Kozdon (jekozdon@nps.edu)
--- a/src/p4est.c
+++ b/src/p4est.c
@@ -201,6 +201,19 @@ p4est_new_ext (sc_MPI_Comm mpicomm, p4est_connectivity_t * connectivity,
                p4est_locidx_t min_quadrants, int min_level, int fill_uniform,
                size_t data_size, p4est_init_t init_fn, void *user_pointer)
 {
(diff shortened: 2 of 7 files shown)


cburstedde:
Jeremy, sounds good. Comments:
1. Can you please commit a doc/author_kozdon.txt file to put your code under FreeBSD.
   And please be sure that you are legally safe to state this.
2. Can you provide full doxygen comments for all parameters in the .h files?
   Btw. p4est is an [in,out] parameter.

jkozdon:
Done. Used CC0 since as a federal employee this is most appropriate.

cburstedde:
Thanks a lot! Nice feature.
I've settled on p4est_reset instead of p4est_destroy_ext.
Can you please give it a try on feature-p4est-init-ext?

cburstedde:
> Done. Used CC0 since as a federal employee this is most appropriate.

I said I'd not use the name destroy_ext -- but maybe that's what it should be
called.  Especially since it's in the extended.h file.  I'll ponder this for a
couple days.

cburstedde:
I've pondered this for a couple more days and am reluctant to make it official, mostly for the naming collision with p4est_init. Also, the p4est3 API may be mostly based on opaque pointers, which is the reverse direction of development. I'll make it a wontfix for now, but should we choose to use this functionality internally I'll make sure to try to expose your work somewhere.

Sorry for the work I've made you do and thanks again!

PR #43, opened 2018-06-01
Application stub files

jkozdon:
The idea of this pull request is to allow a user to add functionality to the p4est library without having to re-bootstrap after pulling a tarball or patching existing p4est files. For instance, a user may want to add accessor functions to p4est structs to linking with a second language.

Right now my workflow for doing this is patch p4est source files prior to build (so that I can just grab the tarball off the web), but this means that I need to create a new patch file every time I want to add a feature which is a bit clumsy.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -16,7 +16,8 @@ libp4est_installed_headers += \
         src/p4est_points.h src/p4est_geometry.h \
         src/p4est_iterate.h src/p4est_lnodes.h src/p4est_mesh.h \
         src/p4est_balance.h src/p4est_io.h \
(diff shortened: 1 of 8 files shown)


cburstedde:
Not a bad idea. What about renaming from _app to _empty?

cburstedde:
Thanks again, I've added another commit to use our header conventions.
I've also run the ./p4estindent script on each modified files and made sure it's C/C++ safe.
If this works for you, I'll merge asap.

cburstedde:
> Fine by me. Changed.

Merged into develop -- any comments?

jkozdon:
This is all fine by me. Thanks!

PR #81, opened 2020-07-22
fix a couple of doxygen warnings

tjhei:
It might make sense to use function parameter names like "forest" instead of "p4est" or "p8est" if you prefer. This just adjusts the documentation.


cburstedde:
Thanks! I've merged this into the develop branch, bypassing this pull request. It will become part of master eventually.

PR #87, opened 2020-11-18
add version functions

ranocha:
This PR is similar to https://github.com/cburstedde/libsc/pull/24 and https://github.com/cburstedde/libsc/pull/25.
- I implemented functions returning the full version as string and the major/minor/point versions as `int`s.
- I added corresponding tests and verified that they pass locally.
- I added a license statement for my contributions in `docs`.
- I ran `p4estindent` and discarded changes to `__attribute__`.

### DIFF
--- a/doc/author_ranocha.txt
+++ b/doc/author_ranocha.txt
@@ -1,0 +1,1 @@
+I place my contributions to p4est under the FreeBSD license. Hendrik Ranocha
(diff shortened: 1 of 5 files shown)


cburstedde:
After I update libsc, you can call SC_TOSTRING from the p4est code.  Let's make libsc be fine with v7 and v8.9 type version styles first and then carry it over to p4est.

ranocha:
I've updated the code to use `SC_TOSTRING` from https://github.com/cburstedde/libsc/pull/26. All tests pass locally for me when using that branch of `libsc`.

Issue #93, opened 2021-02-18
Expand p4est_connectivity_reorder


mugolotti:
Hello,

I would like to make a recommendation for a possible p4est enhancement that would help be extremely helpful in speeding up our code. 
Hers is the description.

**Motivation**
I am working on a high-order CFD solver that uses p4est for feature-based AMR.
The solver is capable of handling high-order elements and follows the following set of operations for reading the connectivity of the elements from the grid file:

- PASS1: read the number of vertices and number of trees from the grid file. Then create p4est connectivity using `p4est_connectivity_new`.
- PASS2: read the nodes coordinates and element connectivities (very vertices of the element, i.e. "low-order" nodes") from the grid file. Then complete the connectivity using `p4est_connectivity_complete`.
- Reorder the connectivity using `p4est_connectivity_reorder`.
- Create p4est using `p4est_new_ext` and providing a callback function that, for each element, reopens the grid file (PASS3) and reads the remaining "high-order" nodes, and stores them in `p.user_data`.

**Description**
The PASS3 consists of finding a specific tree in the grid file and reading high-order node IDs. As of now, the only way to find that specific tree in the grid file is to skim the file and compare the first 4 node IDs (or 8 in 3D) of the tree to find with the first 4 node IDs (or 8 in 3D) of each element in the grid file. Evidently, this is a very time-consuming approach.

Alternatively, we could bookkeep the location of each element in the grid file (for instance line number for ASCII grid files or the bytes count for binary files) during PASS2 and save this data in an array ( `ElemLoc`). 
Then, during the PASS3, we can fast forward to the exact location of the tree in the grid file by using `ElemLoc[tree_id]`.
However, we lose track of the proper location of the trees in the array _ElemLoc_ after the trees are reordered by calling `p4est_connectivity_reorder`.

**Proposed solution**
Threfore, we propose to introduce a new function called `p4est_connectivity_reorder_ext` that carries out the same operations of the current `p4est_connectivity_reorder` function except that reurns the `newid` array.
Furthermore, the function `p4est_connectivity_reorder` will be recasted as follow:

`void`
`p4est_connectivity_reorder (sc_MPI_Comm comm, int k,p4est_connectivity_t * conn,p4est_connect_type_t ctype)`
`{`
 &nbsp;`     sc_array_t         *newid;`
 &nbsp;`     newid = p4est_connectivity_reorder_ext (comm, k, conn, ctype);`
 &nbsp;`     sc_array_destroy (newid);`
`}`

[...]

PR #95, opened 2021-03-01
Extended reorder function.

mugolotti:
# Extended reorder function.

Following up on issue #93.

Proposed solution:

Added a new function called `p4est_connectivity_reorder_ext` that carries out the same operations of the current `p4est_connectivity_reorder` function but it accepts one more IN/OUT argument `newid` array.
Furthermore, the function `p4est_connectivity_reorder` will be recast to allow standard use.

### DIFF
--- a/doc/author_ugolotti.txt
+++ b/doc/author_ugolotti.txt
@@ -1,0 +1,1 @@
+I place my contributions to p4est under the FreeBSD license. Matteo Ugolotti (mugolott@uwyo.edu)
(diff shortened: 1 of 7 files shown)


cburstedde:
Thanks a lot! A couple suggestions:

 * Change to `sc_array_new`, which will use a count of 0 on input.
 * Add assertions on `newid` being not `NULL` and having the correct element size.
 * Add to the newid documentation about pre- and postconditions (non-NULL, element size and count).

mugolotti:
Thanks for the recommendations. 

I think I made the modifications you asked me to do. 
Please, let me know if those are ok or if I need something else.

mugolotti:
Yes, your point makes sense. I should have caught that as well.
I updated the code and pushed the commit. 
Let me know if you have other thoughts. 
Thanks

Issue #94, opened 2021-02-25
VTK Sequential write


ligazetom:
**Description**
VTK file writes are all parallel at the moment. Which is nice, but I have some concerns about file server not being able to handle requests from multiple processes. So I wanted to implement/force sequential write of these files. Problem is the only exposed function for writing cell data in p4est_vtk.h is "p4est_vtk_write_cell_dataf" which takes variadic arguments. Since I would like to wrap this sequential functionality to another function, it has to support variadic arguments as well, but then I can pass them forward only as va_list. 

In p4est_vtk.h there is another function "p4est_vtk_write_cell_data" which would work, however it is not implemented. 
In p4est_vtk.c is function "p4est_vtk_write_cell_datav" which is exactly what I need, but it is not exposed.

Obviously I can do whatever I need since I have the source code, the thing is if I provide someone else with only my project source, they have to make the change to the library as well, which is not ideal. 

EDIT: The quickest fix would be to make copy of "p4est_vtk_write_cell_dataf" function that takes va_list already.

ligazetom:
Since apparently I do something very wrong, I am not able to compile when anything is added in p4est_vtk.c file.

cburstedde:
Thanks, those are all good points. If you'd like to make a PR against the develop branch, that would be great.

Exposing `p4est_vtk_write_cell_datav` sounds good as well as implementing `p4est_vtk_write_cell_data`. When in doubt about the prototypes, just let me know. Please be generous with doxygen comments.

Doing the same for point data would be really grand.

cburstedde:
Thanks for looking into it!

> I went with copy of `p4est_vtk_write_cell_dataf` where I changed variadic arguments for va_list. I thought about exposing the `p4est_vtk_write_cell_datav`, however there are some checks in `p4est_vtk_write_cell_dataf` which would become unnecessary responsibilities for the user.

I'd prefer not duplicating any code. Just exposing `_cell_datav` and adding a bit more documentation on the `va_list` contents seems fine to me.

> However I cannot think of a good name for this "overload", so if you agree with this solution and can think of something better than `p4est_vtk_write_cell_dataf_va_list` I'd gladly create PR.

Exposing the `_cell_datav` function, we won't need another name. :)

> You mentioned "doubt about prototypes". I didn't want to create new issue for something like this, but it's been bothering me for a while. I don't understand why none of refine callbacks (regular or _ext) are getting `void *user_data`.

[...]

PR #96, opened 2021-03-08
Exposing 'p4est_vtk_write_cell_datav' function

ligazetom:
# Exposing `p4est_vtk_write_cell_datav` function for users.

Following up on issue #94 

Proposed changes:
Exposed already implemented function `p4est_vtk_write_cell_datav` for writing vtk cell data using va_list instead of only variadic arguments.

Added (I hope) clear enough doxygen documentation on usage.

Useful for wrapping possibilities, since variadic arguments cannot be passed any other way than va_list.


ligazetom:
Sorry for this one. My first `ever`. I was trying something and suddenly it added my second commit as well so eh..

cburstedde:
You can push or force push to your PR branch and thus update the code. Not nice though, agreed.

PR #97, opened 2021-03-08
Exposed 'p4est_vtk_write_cell_datav' function for users

ligazetom:
# Exposed `p4est_vtk_write_cell_datav` function

Following up on issue #94

Proposed changes:
Exposed already implemented function `p4est_vtk_write_cell_datav` for writing vtk cell data using va_list instead of only variadic arguments.

Added (I hope) clear enough doxygen documentation on usage.

Useful for wrapping possibilities, since variadic arguments cannot be passed any other way than va_list.

### DIFF
--- a/doc/author_krasnansky.txt
+++ b/doc/author_krasnansky.txt
@@ -1,0 +1,1 @@
+I place my contributions to p4est under FreeBSD license. Juraj Krasnansky (xkrasnanskyj@stuba.sk)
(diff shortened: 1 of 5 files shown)


ligazetom:
I'd do the same for the point data, but as of now, I have no way and time to test it, so I don't want to commit something that I do not know is working.

ligazetom:
>     * Would you mind documenting (if briefly) all of the input parameters?
Hah, I copied the `point` version, did not even noticed, even those few descriptions were wrong.

ligazetom:
> P.S. Present tense in commit messages is preferred by many. No big deal.

I'm solo on most projects so this is much appreciated.

Issue #98, opened 2021-03-08
p4est_ghost_expand expandable/failure


ligazetom:
**Description**
I am in point of my project where I need to make region around my quadrant according to user-specified pixel radius. For that I need to make sure I have all of the quadrants in that radius accessible, so the worst case scenario is having the smaller quadrant (size of 1 pixel) at the partition edge.

I found out that if I call `p4est_ghost_expand` too many times I'll eventually get ASSERT failure for `p4est_ghost_is_valid`. It happens if there are at least 3 partitions and the 1st one tries to grab quadrants from the 3rd. 

Obviously that seems a little off to me, but nevermind, I am able to work around it, however there is no way of knowing how many times I can call expand before it fails. Is there some way I can know the number of possible expands before the whole ghost layer gets invalidated? I am working with images so memory is no issue here at the moment. 

**To Reproduce**
You can run step1 on at least 3 processes and adding ghost layer with for loop in main with `p4est_ghost_expand` call.

ligazetom:
I propose to remove `P4EST_ASSERT (p4est_ghost_is_valid (p4est, ghost));` from `p4est_ghost_expand_internal` and call them inside functions that actually call 'p4est_ghost_expand_internal'. Then adding another function of type `bool` that would return `p4est_ghost_is_valid`. With this I could try expand as much as I need and if I hit a limit I'll do another pass with the magic number.

ligazetom:
I needed this functionality very much and ASAP, so if there is better solution I will be glad to implement that. However I solved it by implementing my own 'p4est_ghost_is_valid' function, which is copy of the old one but every `return 0` is substituted by this:

static void HandleSynchronizedReturn(p4est_t *p4est, int *allFlags, int *okFlag, int mpisize, int mpirank)
{
  int i;
  MPI_Gather(okFlag, 1, MPI_INT, allFlags, 1, MPI_INT, 0, p4est->mpicomm);

  if (mpirank == 0)
  {
    for (i = 0; i < mpisize; ++i)
    {
      *okFlag = allFlags[i];
      if (*okFlag == 0) break;
    }
  }

  MPI_Bcast(okFlag, 1, MPI_INT, 0, p4est->mpicomm);

  if (!*okFlag)
  {
    if (mpirank == 0) P4EST_FREE(allFlags);
  }
}

[...]

PR #99, opened 2021-03-09
Fix accidentally enlarged ghost layer in p4est_ghost_expand

tisaac:
Following up on issue #98.

Proposed changes: fix bug where the ghost layer accidentally wasn't resized correctly if none of the proposed new ghost quadrants was actually new.


cburstedde:
Thanks so much for the test case and the fix!

Issue #105, opened 2021-04-04
Memory leak during VTK write


ligazetom:
**Description**
Memory leak (caught by sc) happens during VTK cell data writing, if you don't write out rank and/or level. I checked versions from master 2.2 as well as from develop branch and it happens in all of them.

This is caused by deallocating variables `locidx_data` and `uint8_data` during normal execution only inside of `if` statements for the respective arguments. That is why you need both, level and rank to be 1. If either one of them is 0, one of the variables won't get deallocated.

**To Reproduce**
This is minimal failing example (edited step1, you need hw32.h as well).

[vtk_minimal_fail.zip](https://github.com/cburstedde/p4est/files/6255187/vtk_minimal_fail.zip)

**Additional information**
I will create PR to fix this, just wanted to create issue first, so anyone else can confirm.

cburstedde:
Good point, will look at PR. Thanks!

Issue #31, opened 2018-02-22
Release 2.0


rubendibattista:
Could you please upload the complete `tar.gz` (sc  and `.configure` included) on p4est.org? Or is it still not ready?

cburstedde:
Yes I'm late on that. Will try to get to it.

rubendibattista:
Thanks @cburstedde. I opened a PR for that.

cburstedde:
I've followed your suggestion -- thanks!

PR #106, opened 2021-04-04
Vtk cell data memory leak

ligazetom:
# Memory leak during writing cell data

Following up on issue #105  .
Leak happened every time you did not write out level and/or rank, since deallocation of always allocated variables during normal execution happened inside of `if` statements.

Proposed changes:
No longer always allocating uint8_data, since only `if (write_level)` statements uses it. Now this array is allocated and deallocated inside of this statement.

Variable `locidx_data` is always deallocated right after the `if (write_rank)` statement. Of course proper deallocations in case something goes wrong are in place.

This PR is a little 'hacky' since it is directly influenced by my previous one and both of them are editing the same piece of code.

### DIFF
--- a/src/p4est_vtk.c
+++ b/src/p4est_vtk.c
@@ -1104,7 +1104,6 @@ p4est_vtk_write_cell_data (p4est_vtk_context_t * cont,
   p4est_topidx_t      jt;
   p4est_locidx_t      il;
 
-  p4est_vtk_context_t *list_end;
   char                vtkCellDataString[BUFSIZ] = "";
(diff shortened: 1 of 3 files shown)


cburstedde:
Thanks, I like this. Just running last checks before merging.

cburstedde:
Hi, thanks, I'm catching one error with CI (see "Actions" tab):

CI for CMake #31

Would you be able to look at it?

ligazetom:
Created PR. Missed one macro.

Issue #103, opened 2021-03-12
Continue CMake build system


cburstedde:
**Description**

The CMake build system is supported on branch prev3-develop, but not yet complete.

**Proposed solution**

 * Always compile the example programs.
 * The archive created by `cpack` is lacking the `.c` source files as well as example and test.
 * The archive created by `cpack` shall include all files required to run `./bootstrap && configure && make  -j check V=0` in the unpacked tree.
 * Generally, the `cpack` archive should be as similar as possible to the `make dist` tarball by automake.

cburstedde:
Additionally:

 * Add a new CI job that runs cpack, unpacks the archive und then runs the autotools chain.
 * Add a new CI job that calls make dist, unpacks and then builds, tests and packs using cmake.

cburstedde:
Another point:

 * The tests on make check run with 2 MPI ranks. Make ctest behave the same if not so.

cburstedde:
A couple more items to align the CMake to the automake build:

 * Currently, make install without explicit prefix installs into sc/local. Change to local.
 * The CMake build creates libp4est.a, libp6est.a and libp8est.a. To align with autotools, it should be all in libp4est.a.

cburstedde:
* The example subdirectory seems to be a separate package. Integrate into main.

scivision:
I think to make "examples" be a full test of the "installed" p4est from a single project instead of two separate projects as I have now would require making CMake be a "superbuild" where a new top-level CMakeLists.txt contains two ExternalProjects--one for p4est, one for p4est examples. 

The reason for examples as a separate project or subproject is to test that CMake packaging is correct--it is a common pitfall to have  problems with linking, compiling or finding the installed library due to something wrong with p4est CMake package. These problems only show up when treating p4est as a separate package upon a separate project or subproject, but are how virtually all non-p4est-dev users will use p4est.

scivision:
To specifically control tests MPI rank, lines like

```cmake
${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} ${Ncpu} $<TARGET_FILE: ...
```

become like (assuming fixed rank 2 is desired)

```cmake
${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} 2 $<TARGET_FILE: ...
```

cburstedde:
[...]

PR #108, opened 2021-04-29
CMake: correct autobuild of libsc

scivision:
previously I directly added libsc to p4est CMake project. This created undesired mixing of project scopes, such that libsc did not install correctly when built with p4est. This PR eliminates that problem and clarifies linking by keeping libsc scope totally separate from p4est in CMake.


cburstedde:
Thanks! Please cf. #103 for continued discussion of cmake.

PR #110, opened 2021-05-13
make single p4est.{so,a} library incl. p6est,p8est like autotools

scivision:
Make a single p4est binary library, incorporating p6est and p8est objects, like autotools


scivision:
The Mac failure may be a CI glitch as it passed on mine: https://github.com/scivision/p4est/actions/runs/837856571

cburstedde:
Thanks, is the test failure above relevant? In general, happy to merge.

PR #111, opened 2021-05-21
Fix CMake build

ligazetom:
# CMake build

Following up on issue with CI for CMake #31

Proposed changes:
Add preprocessor macro.

### DIFF
--- a/src/p4est_vtk.c
+++ b/src/p4est_vtk.c
@@ -1320,7 +1320,10 @@ p4est_vtk_write_cell_data (p4est_vtk_context_t * cont,
       p4est_vtk_context_destroy (cont);
 
       P4EST_FREE (names);
+
+#ifndef P4EST_VTK_ASCII
       P4EST_FREE (locidx_data);
+#endif
 
       return NULL;
     }


cburstedde:
Cool thanks. I haven't compiled ascii VTK since forever.

Btw I like your work. In forestclaw I've written the VTK output to create one giant MPI I/O .vtu file using appended binary data. Would you be interested in backporting something like this to p4est, activated by setting a context property?

ligazetom:
Ah great. This is something I've wanted to do as well, but did not really feel like going that deep. I've just created separate directories for every task and animations.

Since I believe the main logic is already there, it shouldn't be too much of a hassle :) Just point me to the implementation.

cburstedde:
> Ah great. This is something I've wanted to do as well, but did not really feel like going that deep. I've just created separate directories for every task and animations.
> 
> Since I believe the main logic is already there, it shouldn't be too much of a hassle :) Just point me to the implementation.

https://github.com/ForestClaw/forestclaw (Look for the VTK output).

I'm thinking to keep the p4est defaults exactly as is.  Add:

p4est_vtk_context_set_append (int understood as bool, default 0)

to activate the MPI I/O out into one big file.
We'd ignore the P4EST_VTK_BINARY and COMPRESSION macros in this case.
When P4EST_ENABLE_MPIIO is not defined, always assume append = 0
and do not compile the new MPI I/O code.

ligazetom:
I'll have a look in the near future.

PR #114, opened 2021-05-28
Fix `p4est_save_ext` on Windows

### LINKED ISSUE #113: `p4est_save_ext` fails on Windows due to different behavior of `fopen`/`ftell`
**Description**
The function `p4est_save_ext` fails under Windows at the following line:
https://github.com/cburstedde/p4est/blob/3a78fd2d3a45d27a96db108e042838d0da12591c/src/p4est.c#L3527
A similar issue occurs in `p6est_save_ext` at
https://github.com/cburstedde/p4est/blob/3a78fd2d3a45d27a96db108e042838d0da12591c/src/p6est.c#L761
I manually instrumented the code at the problematic locations to show the return value of `ftell`, which gives the number of bytes from the beginning of the file to the file position indicator (see, e.g., [here](https://en.cppreference.com/w/c/io/ftell)). The return value is indeed `0`, indicating an empty file (an actual error would have been indicated by a return value of `-1`). However, by looking at the output file generated until this point, it is clear that the file is indeed non-empty. So what happened?

It took me a while to figure out the reason for this, but it is related due to the different behavior of `ftell` on Windows vs. Linux/macOS:

A few lines before the line in which the error occurs, the output file is opened in "(binary) append" mode using `fopen(filename, "ab")`:
https://github.com/cburstedde/p4est/blob/3a78fd2d3a45d27a96db108e042838d0da12591c/src/p4est.c#L3521-L3527
According to most sources documenting the C function `fopen` (e.g., [here](https://en.cppreference.com/w/c/io/fopen) or [here](https://pubs.opengroup.org/onlinepubs/9699919799/)), opening a file in "append" mode *seems* to mean that the file position indicator is set to the end of the (possibly existing) file to ensure that new data is written at the end. This behavior is used in `p4est.c#L3527` and `p6est.c#L761` to ensure that the previous write operation has succeeded by asserting the return value of `ftell` to be greater than zero.

However, on Windows the behavior of `fopen`/`ftell` differs in a small, but in this case significant, detail ([source](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/ftell-ftelli64?view=msvc-160)):
> If no I/O operation has yet occurred on a file opened for appending, the file position is the beginning of the file.

and

> When a file is opened for appending, the file position is moved to end of file before any write operation.

This causes the checks in `p4est.c#L3527` and `p6est.c#L761` to fail, since the return value of `ftell` after opening is `0`, even though the file has been created and written to correctly before.

**To Reproduce**
An MWE to produce this error is, e.g., as follows:
```c
#include <p4est.h>

int main(int argc, char **argv)
{
  p4est_t            *p4est;
  p4est_connectivity_t *conn;

  conn = p4est_connectivity_new_periodic();
  p4est = p4est_new(0, conn, 0, NULL, NULL);
  p4est_save("testfile", p4est, 0);

  /* Destroy the p4est and the connectivity structure. */
  p4est_destroy (p4est);
  p4est_connectivity_destroy (conn);

  return 0;
}
```
Compiling and running this under Windows yields the following output:
```
Into p4est_new with min quadrants 0 level 0 uniform 1
New p4est with 1 trees on 1 processors
Initial level 0 potential global quadrants 1 per tree 1
Done p4est_new with 1 total quadrants
Into p4est_save testfile
Abort: first file tell
Abort: src/p4est.c:3527
Abort
```

**Additional information**
More information on how to reproduce the erroneous behavior can be found at https://github.com/sloede/p4est-demo. It was reproduced under Windows using a current MSYS2 installation and the `mingw-w64-x86_64-toolchain` package.

sloede:
# Fix `p4est_save_ext` on Windows

Following up on issue #113.

Proposed changes:
After opening the existing files in `p4est_save_ext`/`p6est_save_ext` in "append" mode, an additional call to `fseek` ensures that the file position indicator is moved to the end of the file. On Linux/macOS this should be a no-op, since the file position indicator is already at the end of the file. On Windows, however, this fixes the issues with a zero return value from `ftell` as reported in #113.

P.S.: I realize that the contributing guidelines ask for the creation of an issue first. However, since I had a possible fix already available, I just went ahead to show a possible remedy :-)

Closes #113.

### DIFF
--- a/INSTALL_WINDOWS
+++ b/INSTALL_WINDOWS
@@ -44,9 +44,17 @@ Instructions:
         File sc/src/sc.h line 121:
         #define __attribute__(x)
 
+        File sc/test/test_node_comm.c line 36:
+        #define srandom srand
+        #define random rand
+
         In files src/p4est_algorithms.c, p8est_algorithms.c, p6est.c
         Include definitions of htonl-function
         #define htonl(_val) ( ((uint16_t)(_val) & 0xff00) >> 8 | ((uint16_t)(_val) & 0xff) << 8 )
+
+        File test/test_balance_seeds2.c line 33:
+        #define srandom srand
+        #define random rand
 
     b. Compile the package (MPI is disabled as some changes on signals were made
         and it seems to result in the program crashing when MPI functions are used).
@@ -54,7 +62,7 @@ Instructions:
         cd p4est/
         ./bootstrap
         ./configure --disable-mpi --disable-vtk-build --without-blas
-        ./make
+        ./make LDFLAGS="-no-undefined"
         ./make install
 
   6. Build DLLs suitable for Windows.  Linking .a files to MS VS project
--- a/doc/author_schlottke-lakemper.txt
+++ b/doc/author_schlottke-lakemper.txt
@@ -1,0 +1,1 @@
+I place my contributions to p4est under the FreeBSD license. Michael Schlottke-Lakemper
--- a/src/p4est.c
+++ b/src/p4est.c
@@ -3522,6 +3522,10 @@ p4est_save_ext (const char *filename, p4est_t * p4est,
     file = fopen (filename, "ab");
     SC_CHECK_ABORT (file != NULL, "file open");
 
+    /* explicitly seek to end to avoid bad ftell return value on Windows */
+    retval = fseek(file, 0, SEEK_END);
+    SC_CHECK_ABORT (retval == 0, "file seek");
+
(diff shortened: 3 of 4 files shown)


sloede:
Please note that this PR also fixes the tests `p4est_test_loadsave` and `p8est_test_loadsave` under Windows, which are failing with the current release 2.3.1/2.2 on Windows.

cburstedde:
Thanks! Good catch.

PR #123, opened 2021-10-11
CMake presets cleanup and link improvements

scivision:
* Fix p4est examples library finding/linking
* remove unused CMake presets


cburstedde:
Thanks again!

cburstedde:
CI fails on Windows now. Would you be able to looki into it?

Issue #128, opened 2021-12-17
Bug in `p6est_lnodes` comment?


jkozdon:
Am I correct in that this comment should also say that the edges are also differently numbered.

https://github.com/cburstedde/p4est/blob/45ea68a91d11ee15172bc8e6336ed62c91f09cf4/src/p6est_lnodes.h#L40-L41

cburstedde:
Ok, while we're at it, let's discuss whatever remains to do in #129.

PR #129, opened 2022-01-13
Update some p6est docs

jkozdon:
In calls to `replace_fn` the number of out layers should be set to `4` when horizontal coarsening. See docs for `p6est_replace_t` in `src/p6est.h`.

### DIFF
--- a/src/p6est.h
+++ b/src/p6est.h
@@ -174,7 +174,9 @@ typedef struct p6est
                                              never touched by p4est */
   p6est_connectivity_t *connectivity;   /**< topology of sheet, not owned. */
   p4est_t            *columns;  /**< 2D description of column layout
-                                     built from \a connectivity */
+                                     built from \a connectivity.
+                                     \note columns->p.user_data
+                                     cannot be used. */
   sc_array_t         *layers;   /**< single array that stores
                                      p2est_quadrant_t layers within columns */
   sc_mempool_t       *user_data_pool;   /**< memory allocator for user data */
@@ -330,6 +332,8 @@ p6est_t            *p6est_new (sc_MPI_Comm mpicomm,
  *
  * \return This returns a valid forest.  The user must destroy the
  * connectivity for the new p6est independently.
+ *
+ * \note p4est->p.user_data is not retained.
  */
 p6est_t            *p6est_new_from_p4est (p4est_t * p4est,
                                           double *top_vertices,
--- a/src/p6est_lnodes.h
+++ b/src/p6est_lnodes.h
@@ -37,12 +37,12 @@ SC_EXTERN_C_BEGIN;
  * for the types of hanging faces that occur in a p6est.  Please see the
  * documentation for p8est_lnodes */
 
-/* The only other differece is in the numbering of nodes, edges, and faces.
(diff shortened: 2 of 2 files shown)


jkozdon:
cc @tisaac

jkozdon:
Not sure if this is the right branch or not for this fix.

src/p6est.c
```diff
@@ -1636,7 +1636,7 @@ p6est_replace_column_join (p4est_t * p4est, p4est_topidx_t which_tree,
       }
     }
     if (replace_fn != NULL) {
-      replace_fn (p6est, which_tree, P4EST_CHILDREN, 1, outgoing, q, 1, 1,
+      replace_fn (p6est, which_tree, P4EST_CHILDREN, 4, outgoing, q, 1, 1,
```
cburstedde, line 1639: Good point, but having Toby confirm may still be the safest. Let's make this P4EST_CHILDREN for clarity.
Testing the change, does everything check out, i.e., the pointer to the outlayers contains 4 of them and they are valid?
While you're at it, if you have anything to add to the documentation of the replace function type please feel free.

jkozdon: Done.

jkozdon:
> Testing the change, does everything check out, i.e., the pointer to the outlayers contains 4 of them and they are valid?

Based on the code above the correction, `q` is of size `P4EST_CHILDREN` and all the entries are initialized.

- https://github.com/cburstedde/p4est/blob/prev3-develop/src/p6est.c#L1574
- https://github.com/cburstedde/p4est/blob/prev3-develop/src/p6est.c#L1601
- https://github.com/cburstedde/p4est/blob/prev3-develop/src/p6est.c#L1632

jkozdon:
closes #128

tisaac:
I'm sorry I'm late getting this!

This change seems wrong to me, but it's a bugfix so there must be something I don't understand.

At the call site `replace_fn` should be used to update the data for a change of the following kind:

```
+-----+-----+           +-----------+    
|\     \     \          |\           \
| +-----\-----+         | \           \
+ |\     \     \   =>   +  \           \
 \| +-----+-----+        \  +-----------+
  + |     |     |         \ |           |
   \|     |     |          \|           |
    +-----+-----+           +-----------+
```

The description is confusing and should be updated because "outgoing" means leaving the mesh (left) and "incoming" means new and joining the mesh (right), but controlling for that the number of new layers is 1, not 4.

jkozdon:
> This change seems wrong to me, but it's a bugfix so there must be something I don't understand.

It's also very possible I have misunderstood the expected behavior.

tisaac:
Assuming you have data attached to the layers in the columns, `replace_fn` should be used to initialize data for incoming layers from the data in the outgoing layers that occupy the same location in the mesh.  I guess my questions are

- Is that what you were expecting?
- What kind of bug did you experience?

jkozdon:
Honestly, I am not sure.

Relooking at this I did a few months ago I am not sure why I thought that outlayers should be 4.

I've removed that commit. The comment fixes might still be useful?

cburstedde:
Happy to merge if @tisaac signs off. Thanks for keeping at it!

PR #137, opened 2022-04-10
example for using p4est_search_quadrant to count quadrants

gghosh8495:
# Add an example for using the search_reorder function

Proposed changes: Add an easy-to-read example files that calls search_reorder with various options.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -77,6 +77,7 @@ include example/steps/Makefile.am
 include example/tetgen/Makefile.am
 include example/timings/Makefile.am
 include example/balance_seeds/Makefile.am
+include example/quadrant_counter/Makefile.am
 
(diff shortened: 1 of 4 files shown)


gghosh8495:
Whoops! I made a mistake in the commit comment. It should be p4est_search_reorder instead of p4est_search_quadrant.

cburstedde:
Thanks! And sorry for coming back to this only now. I have a couple suggestions to complete this example:

 * Call the directory example/search and base this PR against the branch prev3-develop.
 * Add an sc_options object to allow for various parameters in the call to search_reorder.

Does this make sense?

cburstedde:
An options object may be more than we need, just a connectivity name and a level on the command line seem fine (for a choice of say three connectivities in either dimension).

PR #138, opened 2022-06-08
Add generic neighbor transformations and use them to fix periodic coordinates for plex

tisaac:
There are four parts to this PR

# 1. Small compiler warning / indent fixes for p6est

Self-explanatory.

# 2. Add operations on pure coordinates

- `p4est_coordinates_compare()`
- `p4est_coordinates_transform_face()`
- `p8est_coordinates_transform_edge()`

From the commit messages:

> It is insufficient just to transform nodes because nodes are restricted to being at valid quadrant corners, so not all coordinates can be represented by nodes.
>
> Likewise, the coordinates can be sorted in Morton order even without a level attached, which is useful for canonicalizing points between trees, so we add this capability.

# 3. Add generic neighbor transformations

- typedef `p4est_neighbor_transform_t`
- `p4est_connectivity_get_neighbor_transforms()`
- `p4est_neighbor_transform_coordinates()` and `p4est_neighbor_transform_coordinates_reverse()`
- `p4est_neighbor_transform_quadrant()` and `p4est_neighbor_transform_quadrant_reverse()`

From the commit messages:

> Add a neighbor transform that generalizes all other inter-tree transformation objects.
>
> - This makes it much easier to write loops over all trees that touch a {face, edge, corner}
> - The data structure, while not as compressed as the others, is more interpretable: each transformation is clearly composed from shifts, permutations, and sign changes.
> - A test is added to make sure the generic neighbor transformation does the same thing as the individual transformations for every interface of every named connectivity.

# 4. Fix vertex coordinates calculations in `p4est_get_plex_data()`

From the commit messages:

> Guarantee that all cells agree about the coordinates of a vertex in plex data
>
> Use newly added neighbor transformations to canonicalize the (tree, qcoords) of each vertex and then apply qcoords_to_vertex to determine the coordinates.
> A periodic connectivity is added test_plex2.c to exercise this capability.

### DIFF
--- a/src/p6est.c
+++ b/src/p6est.c
@@ -162,8 +162,8 @@ p6est_connectivity_memory_used (p6est_connectivity_t * conn)
 {
   return
     p4est_connectivity_memory_used (conn->conn4) +
-    conn->top_vertices == NULL ? 0 :
-    (conn->conn4->num_vertices * 3 * sizeof (double));
+    ((conn->top_vertices == NULL) ? 0 :
+     (conn->conn4->num_vertices * 3 * sizeof (double)));
 }
 
 int
@@ -289,7 +289,7 @@ p6est_connectivity_save (const char *filename, p6est_connectivity_t * conn)
 }
 
 p6est_connectivity_t *
-p6est_connectivity_load (const char *filename, size_t * bytes)
+p6est_connectivity_load (const char *filename, size_t *bytes)
 {
   int                 retval;
   size_t              bytes_in;
@@ -761,7 +761,7 @@ p6est_save_ext (const char *filename, p6est_t * p6est,
     SC_CHECK_ABORT (file != NULL, "file open");
 
     /* explicitly seek to end to avoid bad ftell return value on Windows */
-    retval = fseek(file, 0, SEEK_END);
+    retval = fseek (file, 0, SEEK_END);
     SC_CHECK_ABORT (retval == 0, "file seek");
 
     /* align */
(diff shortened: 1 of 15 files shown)


cburstedde:
Thanks! I'll need into next week to look at this. I'm trusting that it remains legal and supported to have a periodic tree corner for which each tree has different vertex coordinates.

tisaac:
> Thanks! I'll need into next week to look at this. I'm trusting that it remains legal and supported to have a periodic tree corner for which each tree has different vertex coordinates.

Yes, this doesn't change what is legal in a connectivity, only what is canonically decided to be the coordinates associated with points on the boundary between trees with differing coordinates, and only in the plex interface.

src/p6est.c
```diff
@@ -162,8 +162,8 @@ p6est_connectivity_memory_used (p6est_connectivity_t * conn)
 {
   return
     p4est_connectivity_memory_used (conn->conn4) +
-    conn->top_vertices == NULL ? 0 :
-    (conn->conn4->num_vertices * 3 * sizeof (double));
+    ((conn->top_vertices == NULL) ? 0 :
+     (conn->conn4->num_vertices * 3 * sizeof (double)));
```
cburstedde, line 166: Good catch.

cburstedde:
Sorry it took me a long time to get back. Would you think this change may degrade the runtime of p4est_quadrant_compare?

Other than that, I'm fine with everything.

tisaac:
I took a look at the assembly that was generated before and after, and this is my understanding of the difference:

- If the coordinates are equal, the operations are essentially the same.
- If the coordinates are not equal, the new version has one additional `je` comparison of the outputs of `p4est_coordinates_compare()` to zero.

I have not compared the timings.

cburstedde:
Thanks for looking into it, T.

PR #139, opened 2022-06-15
Fix uninitialized data read in repartition

tisaac:
Avoid accessing begin_send_to when it is uninitialized
    
The new loop does not necessarily initialize the 'rank' entry of begin_send_to, so we have to check num_send_to[rank] to see if its necessary.

This bug is causing valgrind errors in PETSc's use of p4est.


cburstedde:
Thanks @tisaac! @tim-griesbach, you mentioned that you may want to add comments / further fixes?

cburstedde:
Thanks again. Any more additions for this one?

tim-griesbach:
> @tim-griesbach, you mentioned that you may want to add comments / further fixes?

Since you already added my commit there are no further fixes that I want to add.
https://github.com/cburstedde/p4est/pull/139/commits/2734b061ba8cc0f11bb1c540f683e7d4c31be677 fixes exactly the same problem as @tisaac described it, but in an other place in the code. valgrind showed for the lines of code that I fixed in https://github.com/cburstedde/p4est/pull/139/commits/2734b061ba8cc0f11bb1c540f683e7d4c31be677 only an error after I printed  `begin_send_to[rank]` at the respective place.

src/p4est_algorithms.c
```diff
@@ -3383,8 +3383,10 @@ p4est_partition_given (p4est_t * p4est,
   last_tree =                   /* same type */
     SC_MAX (last_local_tree, new_last_local_tree);
   my_base = (rank == 0) ? 0 : (global_last_quad_index[rank - 1] + 1);
-  my_begin = begin_send_to[rank] - my_base;
-  my_end = begin_send_to[rank] + num_send_to[rank] - 1 - my_base;
+  my_begin = ((to_begin_global_quad <= rank && to_end_global_quad >= rank) ?
```
tisaac, line 3386: In the case `my_begin > my_rank`, don't we have that `to_begin_global_quad = rank` and `to_end_global_quad = rank`, but `begin_send_to[rank] = -1`?  I don't think this handles all cases correctly.

tim-griesbach: > In the case my_begin > my_rank, don't we have that to_begin_global_quad = rank and to_end_global_quad = rank, but begin_send_to[rank] = -1?

Yes, I agree that we have `begin_send_to[rank] = -1` for `my_begin > my_end` and that this case is included in the conditional expression that I added in this commit.
However, the exact new values of `my_begin` and `my_end` are only relevant for non-empty ranks. If the rank is empty, we only need to ensure that `if (from_begin <= my_end && from_end >= my_begin) {` in line 3413 is evaluated to false. Therefore, it should be enough that in cases where we expected `begin_send_to[rank] = -1` in the old version of the code, it still holds `my_end > my_begin` but this is true since for `begin_send_to[rank] = 0` we obtain `0 >= -my_base = my_begin > -my_base - 1 = my_end`.
Since it also holds `0 <= from_begin <= from_end` the related if-statement is evaluated to false and we should obtain the same behavior of the code as before.
I agree that the conditional expression may cause confusion and therefore I added a comment (https://github.com/tim-griesbach/p4est/commit/85c61140dcecec2e85f8551db11c900ba17d00a8). I also wrote code that is executed in the debug mode that verifies for the tests based an on the old implementation that we get same evaluation for the above mentioned if-statement if you want I could also this code.

cburstedde: Thanks, @tim-griesbach with you assigned you should have push permission now. Could you please add your comment commit to this PR? Also, even better than comments are assertions. Please add a few to hard-prevent the corner cases.

tim-griesbach: > Thanks, @tim-griesbach with you assigned you should have push permission now. Could you please add your comment commit to this PR?

I got a "permission denied" error and according to this link (https://stackoverflow.com/questions/20928727/adding-commits-to-another-persons-pull-request-on-github) it seems to require write access to your p4est repository to add a commit to a PR that was not authored by me.

> Also, even better than comments are assertions. Please add a few to hard-prevent the corner cases.

I will add assertions to hard-prevent corner cases.

PR #141, opened 2022-07-05
CMake: unify CI with libsc, enhance for Windows

scivision:
CI: unify with libsc. test many more compiler configurations as well as shared/static lib


cburstedde:
Thanks!

PR #142, opened 2022-07-07
update p4est cmake

gghosh8495:
# Title

Following up on issue # .

Proposed changes:


gghosh8495:
I am unsure of the #cmakedefine comments, I thought it was written as ' define to 1 ' but it looks different. Just to confirm, we are supposed to write it as ' define to 1 ' right ?

cburstedde:
Yes, this time it's ok to align the comments with the usual form "Define to 1" (when in doubt, stay close to the comments created by autoconf).

Can you please enable and check the version test with cmake?

gghosh8495:
@cburstedde Do we need to add an equivalent (P4EST_CXX \"${MPI_CXX_COMPILER}\") in libsc ? This exists in the config.cmake file.

gghosh8495:
@cburstedde I think all cmake files in p4est and libsc have all that is required. There will be a diff because both requires check which are specific to the package.

gghosh8495:
@cburstedde Ready to merge!

cburstedde:
Thanks!

PR #144, opened 2022-08-15
Add high-order visualization to p4est_vtk

gseastream:
# Add high-order visualization to p4est_vtk

Proposed changes:
- Add high-order visualization ability
  - New function in `p4est_vtk.c` called `p4est_vtk_write_header_ho`, takes coordinates from user and number of nodes in one dimension
  - Minor changes to other functions to allow arbitrary number of points per cell
- Added `p4est_step5.c` example illustrating new capability

I'm quite new to this library, so I definitely may have done things in 'not the right' way. I'm happy to make any changes.


gseastream:
Here are the results of the new step5 example, in both 2D and 3D using both low-order (old code, still available in this PR) and high-order (new). You can see it is a simple example with one level of refinement, with densely varying data within each cell (sum of sin, cos, and sqrt functions). The low-order view misses this data almost completely, but the high-order view gives you a complete picture.

| | 2D | 3D |
| ------------- | ------------- | ------------- |
| **Low Order** | old_2d | old_3d |
| **High Order** | new_2d | new_3d |

cc: @arash-fathi-xom

ackirby88:
@gseastream This is a nice capability so thanks for doing the implementation!

However, there is a bug in your test case implementation. If you try using less than 8 plotting points, e.g., 4, and use the 2D executable, you should get a segfault error. The issue is in getting the vertex coordinates via the p4est function which provides (x,y, AND z=0.0) even for 2D simulations. So the array you pass in is only allocated for DIM*P4EST_CHILDREN and the function for filling the quadrant nodes goes past your memory buffer.

Second, maybe a performance suggestion (not sure if Carsten agrees), but when passing in the `sc_array_t **positions`, the memory has to jump the entire grid to get each coordinate x,y, and z. My suggestion is to stack the coordinates (x,y,z) at each node and pass in `sc_array_t *positions` instead, which is of size `P4EST_DIM*nnodes`.

gseastream:
@ackirby88 thank you for taking a look!

I'm not sure how you're getting the error in the test case. When I do 4 or 5 points it works as expected. I assume you are talking about changing the number of points in each direction from 8 to e.g. 4. To clarify, this is done by changing line 44 in `p4est_step5.c` to the following:

```c
#define STEP5_NNODE_1D 4
```

My whole process in case it matters:

- I cloned my fork from scratch
- checked out my branch
- `git submodule init` and `git submodule update` for the sc dependency
- load mpich/3.4 module
- `./bootstrap`, `mkdir build`. `cd build`
- `../configure CFLAGS="-O0 -g -Wall -Wextra" --enable-debug --enable-mpi`
- `make`
- run `example/steps/p4est_step5`

ackirby88:
`#define STEP5_NNODE_1D 4`: Correct for me to generate a bug on my system.
But it is probably a coincidence that it only crashes when changing that parameter for me.

The bug is persistent for all 2D runs. The issue is with call `step5_qcoord_to_vertex_all` and your `vxy_all[]` input array.
The function `p4est_qcoord_to_vertex` always returns 3 coordinates, not `P4EST_DIM`, so when you get vertex 3 for the 2D simulation, the memory runs out of bounds.

The function signature for `p4est_qcoord_to_vertex`:
```
void                p4est_qcoord_to_vertex (p4est_connectivity_t *
                                            connectivity,
                                            p4est_topidx_t treeid,
                                            p4est_qcoord_t x,
                                            p4est_qcoord_t y, double vxyz[3]);
```

gseastream:
Oh, I see now. I'll have to fix this bug in the code it came from too.

What happens is, in the following, I send parts of `vxy_all` to `p4est_qcoord_to_vertex`, but each part is only expected to be of size 2 when it needs to be of size 3 regardless of 2D/3D. You probably get a segfault when it tries writing the very last piece to one index beyond the length. I'm probably getting lucky and not triggering a problem writing to that location in memory.

https://github.com/cburstedde/p4est/blob/05afdfe4fb6061a11765bc2a460542199f366482/example/steps/p4est_step5.c#L200-L219

cburstedde:
> What happens is, in the following, I send parts of `vxy_all` to `p4est_qcoord_to_vertex`, but each part is only expected to be of size 2 when it needs to be of size 3 regardless of 2D/3D. You probably get a segfault when it tries writing the very last piece to one index beyond the length. I'm probably getting lucky and not triggering a problem writing to that location in memory.
> 
> https://github.com/cburstedde/p4est/blob/05afdfe4fb6061a11765bc2a460542199f366482/example/steps/p4est_step5.c#L200-L219

The output of `_to_vertex` is always a 3-tuple, even in 2D. Please check out the source of the function.

gseastream:
@ackirby88 @cburstedde thank you for the feedback. I've gone ahead and fixed the bug and changed positions to be a 1D array

cburstedde:
Thanks; I'm not getting to do a review today but I'd be grateful if you turned the `//` comments into C-style comments. Please also run `p4estindent` on the file if not done already. How's valgrind reacting to the latest code?

gseastream:
Fixed the comments and ran p4estindent (somehow I think I did it wrong before, that's why it wasn't indented originally).

For valgrind, I get 0 errors, but I'm not sure exactly how to run it or interpret it. I didn't get any errors before fixing the segfault above, either.

cburstedde:
> Any updates on this?

I've been a bit swamped, but now this is first on my list.  Thanks!

cburstedde:
Thanks again, I'm about to merge after some minor tweaks!

Can I ask you to add doxygen comments to the .h files about your HO header functions?
Doesn't need to be wordy. I'll be quick merging thereafter, promise.

gseastream:
Done!

cburstedde:
Thanks! Now I've tightened the loworder memory use by `#define STEP5_NNODE_1D 2`.
Everything is still valgrind clean for loworder, but I'm not sure I didn't break any of your assumptions.

highorder-loworder
Would you be able to double-check that the loworder code behaves as intended?

gseastream:
Hi Carsten, just finished reading through the changes. The intent of the step5 example was to demonstrate how low-order visualization does not express dense patterns in the underlying data. Thus, the code was written so that the data in each element is always 8x8x8, but if STEP5_HO is not defined, we lose that illustration. Your changes indeed save memory, but it's like saying "of course" the low-order image looks bad, the data itself is bad too.

I think I'd rather keep STEP5_NNODE the way I had it, where we always construct a dense underlying data (like we have in our own applications), but allow the user of this example to explore how that same data is visualized by the old and new methods.

The other 'static' changes and the 'if 1' look good. Thanks!

cburstedde:
> Hi Carsten, just finished reading through the changes. The intent of the step5 example was to demonstrate how low-order visualization does not express dense patterns in the underlying data. Thus, the code was written so that the data in each element is always 8x8x8, but if STEP5_HO is not defined, we lose that illustration. Your changes indeed save memory, but it's like saying "of course" the low-order image looks bad, the data itself is bad too.
> 
> I think I'd rather keep STEP5_NNODE the way I had it, where we always construct a dense underlying data (like we have in our own applications), but allow the user of this example to explore how that same data is visualized by the old and new methods.
> 
> The other 'static' changes and the 'if 1' look good. Thanks!

I appreciate the answer.  I'll  roll back that HO change and paste in your
comments above.  Thanks!

gseastream:
Any updates?

PR #145, opened 2022-08-18
Estimation of a derivative in p4est_step3 example

mkirilin:
# Always true condition in the p4est_step3 example
Currently we update a derivative and compare to the same variable which leads to always-true condition.

Proposed changes:
Check if the derivative variable was updated and proceed to the corresponding branch of if-else statement.


cburstedde:
Thanks, please add a proper description and when you like it make it a non-draft pull request.

cburstedde:
@tisaac  what do you think?

PR #146, opened 2022-08-25
cmake: unify with libsc: correct cpack parameters and flatten CMake meta install

scivision:
same as https://github.com/cburstedde/libsc/pull/84


cburstedde:
I'm wondering, is there a way to make the CI run clean with this PR?

cburstedde:
I'm expecting to merge this after integrating with the latest libsc first.

PR #147, opened 2022-08-30
Function for uniform global first quadrant array

tim-griesbach:
# Function for uniform global first quadrant array

Proposed changes:
Implement a function to compute the global first quadrant array for a uniform partition. This function is used in the respective places in p4est. This PR is a preparation for the parallel I/O PR since this function is required for the upcoming code.

### DIFF
--- a/src/p4est_communication.c
+++ b/src/p4est_communication.c
@@ -403,10 +403,8 @@ p4est_comm_global_first_quadrant (p4est_gloidx_t global_num_quadrants,
 
   gfq[0] = 0;
   for (i = 1; i < mpisize; ++i) {
-    gfq[i] = p4est_partition_cut_gloidx ((uint64_t)
-                                         global_num_quadrants, i, mpisize);
+    gfq[i] = p4est_partition_cut_gloidx (global_num_quadrants, i, mpisize);
   }
-
   gfq[mpisize] = global_num_quadrants;
 }
 
--- a/src/p4est_communication.c
+++ b/src/p4est_communication.c
@@ -398,16 +398,14 @@ p4est_comm_global_first_quadrant (p4est_gloidx_t global_num_quadrants,
   int                 i;
 
   P4EST_ASSERT (gfq != NULL);
+  P4EST_ASSERT (mpisize >= 0);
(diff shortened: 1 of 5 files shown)


cburstedde (changes requested):
Thanks, I know this is tedious, but please use the cut function. I'd like to remove the explicit computation from everywhere except inside the cut function.

src/p4est_communication.c
```diff
@@ -398,16 +398,14 @@ p4est_comm_global_first_quadrant (p4est_gloidx_t global_num_quadrants,
   int                 i;
 
   P4EST_ASSERT (gfq != NULL);
+  P4EST_ASSERT (mpisize >= 0);
```
cburstedde, line 401: >= 1 bitte.

tim-griesbach:
I changed the code such that `p4est_partition_cut_uint64` is used inside the loop and I added assertions on `mpisize` and `global_num_quadrants`. I used `p4est_partition_cut_uint64` since it was used in the places where now `p4est_comm_global_first_quadrant` is used. However, we could also use `p4est_partition_cut_gloidx` to save the type cast.

cburstedde:
> I changed the code such that `p4est_partition_cut_uint64` is used inside the loop and I added assertions on `mpisize` and `global_num_quadrants`. I used `p4est_partition_cut_uint64` since it was used in the places where now `p4est_comm_global_first_quadrant` is used. However, we could also use `p4est_partition_cut_gloidx` to save the type cast.

Yes please use the gloidx version.

src/p4est_communication.c
tim-griesbach: > `>= 1` bitte.

Done

tim-griesbach:
> Yes please use the gloidx version.

I pushed the corresponding change.

tim-griesbach:
The CI fails due to:
```
There are not enough slots available in the system to satisfy the 2
slots that were requested by the application:

  /home/runner/work/p4est/p4est/build/test/test_conn_reduce3
```
Given our experiences this should be resolved by a re-run of the Github CI action.

cburstedde (approved):

PR #149, opened 2022-08-31
p4est parallel data file functions

tim-griesbach:
# p4est parallel data file functions

Proposed changes: This PR implements functionality to read and write data in parallel that may be associated with a p4est grid. The corresponding workflow is demonstrated in [`test_io2.c`](https://github.com/tim-griesbach/p4est/blob/4ee84fb5b35e10facd5a679148912fa3fa2d16b4/test/test_io2.c).


cburstedde:
Thanks! Strictly we'd have to double-check that every single update of libsc refers to a version that is an ancestor of the present prev3-develop branch. Is there an automatic/scriptable way to do this, or even a simple set of git commands?

tim-griesbach:
I wrote a bash script to check for the commits that are added by the merge of feature-mpi-write. Hereby, check means that the libsc commit that is associated with the p4est commit is checked with respect to its ancestor relation to libsc/prev3-develop.
 
The script is not optimized in the sense that for each p4est commit the corresponding libsc commit is checked even if the libsc commit was already checked for an other p4est commit. However, the scripts terminates for me in less than 30 seconds and it  does not show any non-ancestor relation.

The script must be executed in the p4est source folder with present .git directory and there must exists a sub folder sc that also contains a .git directory.

```
#!/bin/bash

count=0

for commit in $(git rev-list prev3-develop..feature-mpi-write)
do
  # loop over all commits that feature-mpi-write is ahead of prev3-develop
  # get the corresponding libsc commit
  sc=$(git ls-tree $commit | grep sc | cut -d' ' -f3)
  # extract hash value
  sc_commit=$(echo $sc | cut -d' ' -f1)
  cd sc
  # check ancestor relation (grep with 'not' to see only negativ results)
  if $( git merge-base --is-ancestor $sc_commit prev3-develop ); then
    echo "$sc_commit is an ancestor of libsc/prev3-develop"
  else
    echo "$sc_commit is not an ancestor of libsc/prev3-develop"
  fi
  cd ..
  ((count++))
done

echo "Checked" $count "p4est commits"
```

test/test_io3.c
```diff
@@ -0,0 +1,26 @@
...
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with p4est; if not, write to the Free Software Foundation, Inc.,
+  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+*/
+
+#include <p4est_to_p8est.h>
+#include "test_io2.c"
```
cburstedde, line 26: Missing newline at end of file, please correct.

src/p4est_extended.h
```diff
@@ -44,6 +44,7 @@
 #include <p4est_mesh.h>
 #include <p4est_iterate.h>
 #include <p4est_lnodes.h>
+#include <p4est_io.h>
```
cburstedde, line 47: Would it be sensible to remove this include dependency by forward declaring `struct p4est_file_context`?

src/p4est_io.h
```diff
@@ -68,6 +108,398 @@ p4est_t            *p4est_inflate (sc_MPI_Comm mpicomm,
...
+ *                            by spaces.
+ * \param [out] errcode       An errcode that can be interpreted by \ref
+ *                            p4est_file_error_string and
+ *                            \ref p4est_file_error_class.
+ * \return                    Newly allocated context to continue writing
+ *                            and eventually closing the file. NULL in
+ *                            case of error.
+ */
+p4est_file_context_t *p4est_file_open_create
+  (p4est_t * p4est, const char *filename, const char user_string[15],
```
cburstedde, line 196: The number 15 differs from the above `USER_STRING_BYTES` of 48. Does it make sense to unify these? Also, shall we use a `#define`d value and not the bare number 15 (what is our usual convention)? With the `_ext` version, we provide user_string as `char *`, again, does it make sense to unify?

cburstedde: Please always force the user to input `NUL`-terminated strings, and check that the input length is < 16 (in this case).

src/p4est_io.h
```diff
@@ -68,6 +108,398 @@ p4est_t            *p4est_inflate (sc_MPI_Comm mpicomm,
...
+ *
+ * This function does not abort on MPI I/O errors but returns NULL.
+ *
+ * \param [in] p4est            The forest must be of the same refinement
+ *                              pattern as the one used for writing the file.
+ *                              Its global number of quadrants must match.
+ *                              It is possible, however, to use a different
+ *                              partition or number of ranks from writing it.
+ * \param [in] filename         The path to the file that is opened.
+ * \param [in,out] user_string  At least 16 bytes. The user string is written
```
cburstedde, line 217: If we make this a `#define`, then we can use `\ref` in the documentation. Again, I'd like to decide generally to use `[NUMBER] `or `char *`, but not a mix of the two if that can be avoided.

cburstedde: On both input and output, for strings, allow for the space of the terminating `NUL`, demand it on input and always write it on output.

src/p4est_io.h
```diff
@@ -68,6 +108,398 @@ p4est_t            *p4est_inflate (sc_MPI_Comm mpicomm,
...
+ * \return                    Return the input context to continue writing
+ *                            and eventually closing the file. The return
+ *                            value is NULL in case of error, then
+ *                            it also holds errcode != 0 and the file is
+ *                            tried to close and fc is freed.
+ */
+p4est_file_context_t *p4est_file_write_header (p4est_file_context_t * fc,
+                                               size_t header_size,
+                                               const void *header_data,
+                                               const char user_string[47],
```
cburstedde, line 262: As above, please add space for the terminating `NUL` (so 48 characters) and decide on using a `#define` consistently.

src/p4est_io.h
```diff
@@ -68,6 +108,398 @@ p4est_t            *p4est_inflate (sc_MPI_Comm mpicomm,
...
+ *                                  MPI communicator, and to verify the
+ *                                  global quadrant count found in the file.
+ * \param [in]  filename            Path to parallel file.
+ * \param [in,out] user_string      At least 16 bytes. This array will
+ *                                  be filled with the user string of the
+ *                                  file after a successful call of this
+ *                                  function.
+ * \param [in,out] blocks           After a successful function call this
+ *                                  variable holds an array with a length
+ *                                  corresponding to the number of arrays in the
```
cburstedde, line 443: The number of arrays in the file, or rather the number of elements (maybe not) or data pieces (maybe not) or data blocks (hmm)? Wondering what's the best term.

src/p4est_io.h
```diff
@@ -68,6 +108,398 @@ p4est_t            *p4est_inflate (sc_MPI_Comm mpicomm,
...
+/** Turn p4est_file errcode into a string.
+ * errclass must be a class that is output by \ref p4est_file_error_class.
+ * \param [in] errclass     An errclass that is output by \ref
+ *                          p4est_file_error_class.
+ * \param [in,out] string   At least sc_MPI_MAX_ERROR_STRING bytes.
+ * \param [out] resultlen   Length of string on return.
+ * \return                   sc_MPI_SUCCESS on success or
+ *                           something else on invalid arguments.
+ */
+int                 p4est_file_error_string (int errclass, char *string,
```
cburstedde, line 487: Would it be legale to pass an error code as well as an error class? If yes, please update the docs accordingly. Can we follow the convention used in MPI?

src/p4est_io.h
cburstedde: Let's call each section of the file (i. e. the file metadata, each header, and each field) a "section", officially.

src/p4est_io.h
tim-griesbach: > The number 15 differs from the above `USER_STRING_BYTES` of 48. Does it make sense to unify these? Also, shall we use a `#define`d value and not the bare number 15 (what is our usual convention)?

Yes, I unified the number of bytes of the user string and used the macro for the number of bytes.

>  With the _ext version, we provide user_string as char *, again, does it make sense to unify?

Yes, I used also the macro for number of bytes of the user string at this place.

tim-griesbach: > Please always force the user to input `NUL`-terminated strings, and check that the input length is < 16 (in this case).

This is now also done including assertions using `strlen`

src/p4est_io.h
tim-griesbach: > On both input and output, for strings, allow for the space of the terminating `NUL`, demand it on input and always write it on output.

This is now done for input strings as described above and was already the case for the output strings.

src/p4est_io.h
tim-griesbach: > As above, please add space for the terminating `NUL` (so 48 characters) and decide on using a `#define` consistently.

This is done as described above.

tim-griesbach: > As above, please add space for the terminating `NUL` (so 48 characters) and decide on using a `#define` consistently.

The null-termination of string is done as described above. The defines for the number of bytes of a user string is now used consistently.

src/p4est_io.c
```diff
@@ -469,22 +469,22 @@ check_file_metadata (sc_MPI_Comm mpicomm, const char *filename,
   }
 
   /* check the format of the user string */
-  if (metadata[47] != '\n') {
+  if (metadata[79] != '\n') {
```
cburstedde, line 472: Can we use the `#define`s here? Otherwise the code will break when we change it.

src/p4est_io.c
```diff
@@ -469,22 +469,22 @@ check_file_metadata (sc_MPI_Comm mpicomm, const char *filename,
...
-  user_string[15] = '\0';
+  strncpy (user_string, &metadata[32], P4EST_NUM_USER_STRING_BYTES - 1);
+  user_string[P4EST_NUM_USER_STRING_BYTES - 1] = '\0';
 
   /* check number of global quadrants */
   /* there is no \n at the end of this line */
   metadata[P4EST_NUM_METADATA_BYTES] = '\0';
 
-  if (strlen (&metadata[48]) != 16) {
+  if (strlen (&metadata[80]) != 16) {
```
cburstedde, line 487: The positions 32 and 80 can be changed to be the sum of `#define`s such that we may change the values.

test/test_io2.c
```diff
@@ -142,7 +142,7 @@ write_invalid_files (p4est_t * p4est)
 
     /* invalid3 */
     snprintf (string0, P4EST_NUM_METADATA_BYTES + 1,
-              "%.7s\n%-23s\n%-15s\n%.16ld", P4EST_MAGIC_NUMBER,
```
cburstedde: We could use a stringify macro to make it `"first string" SC_TOSTRING (NUM_BYTES_XYZ) "second string"`.

src/p4est_io.h
```diff
@@ -113,10 +113,10 @@ p4est_t            *p4est_inflate (sc_MPI_Comm mpicomm,
  * The file header is written to the file as string without null-termination
  * (called string*) and is therefore readable in a text editor.
  *
- * File Header (64 bytes):
```
cburstedde: We need to take notes that the documentation depends on bare numbers and change that later.

cburstedde:
General comment: Strings are `NUL`-terminated (`char` of value 0) while pointers may be `NULL` (`void *` of value 0).

src/p4est_io.h
tim-griesbach: Now, there are only error codes in p4est and they can all be handled by `p4est_file_error_string`.

src/p4est_io.h
tim-griesbach: I renamed the data blocks to data sections.

src/p4est_io.h
```diff
@@ -22,13 +22,53 @@
...
+                                                          needs to contain two
+                                                          newline characters and
+                                                          therefore this is the
+                                                          maximal number of pad
+                                                          bytes. */
+#define P4EST_NUM_USER_STRING_BYTES 48 /**< number of user string bytes */
+#define P4EST_NUM_FIELD_HEADER_BYTES (2 + P4EST_NUM_ARRAY_METADATA_BYTES + P4EST_NUM_USER_STRING_BYTES)
+                                     /**< number of bytes of one field header */
+#define P4EST_FILE_COUNT_ERROR -1 /**< All other error codes are defined by MPI or are
+                                     errno. This error code is used to indicate a read
```
cburstedde, line 60: The reference to errno may now be outdated (everywhere in p4est).

src/p4est_to_p8est.h
```diff
@@ -131,6 +142,10 @@
 #define p4est_wrap_leaf_t               p8est_wrap_leaf_t
 #define p4est_wrap_flags_t              p8est_wrap_flags_t
 #define p4est_vtk_context_t             p8est_vtk_context_t
+#define p4est_file_context              p8est_file_context
```
cburstedde, line 145: We usually get away without redefining the names of struct (see rest of p4est).

src/p4est_io.h
tim-griesbach: I changed this comment and checked the rest of the documentation.

cburstedde:
With the test, is there a typo in the filename `invaild2.p8data`?

src/p4est_to_p8est.h
tim-griesbach: If I remove the redefinition I get a compiler error:

```
../p4est/src/p4est_to_p8est.h:145:41: error: invalid application of ‘sizeof’ to incomplete type ‘p8est_file_context_t {aka struct p8est_file_context}’
 #define p4est_file_context_t            p8est_file_context_t
                                         ^
../p4est/src/p4est_base.h:200:65: note: in definition of macro ‘P4EST_ALLOC’
                                                    (n) * sizeof(t))
                                                                 ^
../p4est/src/p4est_io.c:568:53: note: in expansion of macro ‘p4est_file_context_t’
   p4est_file_context_t *file_context = P4EST_ALLOC (p4est_file_context_t, 1);
```

tim-griesbach:
> With the test, is there a typo in the filename `invaild2.p8data`?

Yes, fixed the typo.

cburstedde:
> > With the test, is there a typo in the filename `invaild2.p8data`?
> 
> Yes, I fixed the typo.

Cool. Can you (not necessarily today) use a macro for the file name and follow the test_io_ name convention (see other files)?

What about changing the file extension to `p4d` in accordance with `p4c`, `p4p`?

src/p4est_to_p8est.h
cburstedde: I see. Please solve the problem in the same way as with the VTK context: use an `#ifdef` inside the `.c` file for the redefinition.

src/p4est_io.c
tim-griesbach: Yes, if we introduce additional macros for the entries of the file header. I did this to make `check_file_metadata` independent of the size of the entries of the user header.

src/p4est_io.c
tim-griesbach: Yes, see also the related comment of the review above.

src/p4est_io.c
cburstedde: Thanks, sounds good.

test/test_io2.c
tim-griesbach: Do you prefer the usage of the stringify macro? If I understand your proposed usage of `SC_TOSTRING` correct `snprintf` has the advantage of being explicit on the number of character that are used express the different number of bytes.

cburstedde: We can keep the `snprintf` format strings, but should keep a note that when changing any element lengths, the code needs to be updated. The stringify option would replace 7, 23, 15, 16 by evaluating macros to strings for the same result, with future-proof but more difficult to understand code.

tim-griesbach:
> > > With the test, is there a typo in the filename `invaild2.p8data`?
> > 
> > 
> > Yes, I fixed the typo.
> 
> Cool. Can you (not necessarily today) use a macro for the file name and follow the test_io_ name convention (see other files)?
> 
> What about changing the file extension to `p4d` in accordance with `p4c`, `p4p`?

I changed the file name and the file extension according to your proposition.

cburstedde:
> > > > With the test, is there a typo in the filename `invaild2.p8data`?
> > > 
> > > 
> > > Yes, I fixed the typo.
> > 
> > 
> > Cool. Can you (not necessarily today) use a macro for the file name and follow the test_io_ name convention (see other files)?
> > What about changing the file extension to `p4d` in accordance with `p4c`, `p4p`?
> 
> I changed the file name and the file extension according to your proposition.

Beautiful!

src/p4est_to_p8est.h
tim-griesbach: Ah, I see. The redefinition in the `.c` is now implemented.

test/test_io2.c
cburstedde: Let's call the snprintf format strings good for today and take a note to reconsider them, at the same time that we think about explicit byte counts in the documentation.

cburstedde:
Thanks so much for this massive amount of work!

PR #150, opened 2022-09-07
upload log files as artifacts

gghosh8495:
Upload artifacts for cmake build


cburstedde:
Thanks, looks good. While you're at it, it seems the CMake binary archive uses the same name thrice. What about changing it to mac_archive, linux_archive, windows_archive? Please use an underscore in your names as well.

cburstedde:
I made the suggested change and am now merging.

PR #151, opened 2022-09-08
lock mpisize to 2 for tests

gghosh8495:


cburstedde:
Thanks; I'd prefer to leave Ncpu unchanged, since it's information from the system (see cmake/config.cmake), and rather add a new variable that is 1 or 2 depending on the available number of cores.

gghosh8495:
@cburstedde I added MPIEXEC_NUMPROC_MAX variable which is the same as in libsc, it is only set to 2 if MPI is activated.

cburstedde:
The numprocs_max variable may best be set in config.cmake, where Ncpu is set, since there is an if condition whether the available number of processes is less than 2, in which case numprocs_max should be 1, not 2.

cburstedde:
I've made this change and am now merging.

PR #152, opened 2022-09-20
Use p4est_file_error_cleanup in all clean up macros

tim-griesbach:
# Use p4est_file_error_cleanup in all clean up macros

Proposed changes: There was one macro that did not use `p4est_file_error_cleanup`. This caused errors if one uses OpenMPI instead of MPICH since the related place in the code had a MPICH specifc assumption. This is fixed by using `p4est_file_error_cleanup` in the respective code place.
Beside these changes this PR fixes some typos in the documentation and a compiler warning due to missing type casts.

### DIFF
--- a/src/p4est_extended.h
+++ b/src/p4est_extended.h
@@ -602,7 +602,9 @@ p4est_t            *p4est_source_ext (sc_io_source_t * src,
  */
 p4est_file_context_t *p4est_file_open_read_ext (sc_MPI_Comm mpicomm,
                                                 const char *filename,
-                                                char *user_string,
+                                                char
+                                                user_string
+                                                [P4EST_NUM_USER_STRING_BYTES],
                                                 p4est_gloidx_t *
                                                 global_num_quadrants,
                                                 int *errcode);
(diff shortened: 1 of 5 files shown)


src/p4est_extended.h
```diff
@@ -602,7 +602,9 @@ p4est_t            *p4est_source_ext (sc_io_source_t * src,
  */
 p4est_file_context_t *p4est_file_open_read_ext (sc_MPI_Comm mpicomm,
                                                 const char *filename,
-                                                char *user_string,
+                                                char
+                                                user_string
+                                                [P4EST_NUM_USER_STRING_BYTES],
```
cburstedde, line 607: Wondering about the [FIXEDSIZE] element. Were we not getting rid of this construction?

tim-griesbach: Yes, but I did not added this to this PR. This PR has only the purpose to fix the darwin ci run and smoothing some other places in the code. The code change above was done to have a consistent user string parameter convention.
Do you want that I include the adjustment of the function interfaces already in this PR?

cburstedde: Ah I see. No this can wait, then.

PR #153, opened 2022-09-23
Improved error handling in p4est_io

tim-griesbach:
# Improved error handling in `p4est_io`

Proposed changes:
- Introduce p4est file error codes that are returned by the `p4est_file` functions and can be converted to an error string by `p4est_file_error_string`.
- Remove the dependence on `errno.h` in p4est_io.
- `p4est_file` prefix for static function names in p4est_file.
-  Make `test_io{2,3}.c` partition independent and add more restrictive tests for the error handling.
- Check input data and set the error code accordingly if the input data is invalid. The file context is closed in case of invalidity.
- Update the documentation according to the changes described above.
- Update for new sc_io opening mode names.


cburstedde:
It seems the CI is failing since, at the same time, we would need to upgrade libsc to the latest prev3-develop branch.
I like the fact that this update is not part of this PR. I can myself add one commit to do this. Would that make sense?

tim-griesbach:
Yes, that make sense. I could also add such a commit if you want me to do it.

cburstedde:
> Yes, that make sense. I could also add such a commit if you want me to do it.

Thanks but I think I'll do that just before merging.

src/p4est_io.h
```diff
@@ -58,19 +58,9 @@ SC_EXTERN_C_BEGIN;
...
-
-/** This macro is used for file format errors. */
-#ifndef P4EST_ENABLE_MPIIO
-#define P4EST_ERR_IO EIO /**< File format error code without MPI IO */
-#else
-#define P4EST_ERR_IO sc_MPI_ERR_IO /**< File format error with MPI IO */
-#endif
+#define P4EST_FILE_MAX_GLOBAL_QUAD 9999999999999999 /**< maximal number of global quadrants */
+#define P4EST_FILE_MAX_BLOCK_SIZE 9999999999999 /**< maximal number of block bytes */
+#define P4EST_FILE_MAX_FIELD_ENTRY_SIZE 9999999999999 /**< maximal numeber of bytes per field entry */
```
cburstedde, line 63: Not to be too perfectionist, but would it be easier/safer to use something like (?):
`1000UL * 1000UL * 1000UL - 1UL`

tim-griesbach: Yes, I think it would be easier to read and safer if we use `((10000UL * 10000UL * 10000UL * 10000UL) - 1UL)`

tim-griesbach: I changed the macros to `((10000UL * 10000UL * 10000UL * 10000UL) - 1UL)` and `((10000L * 10000L * 10000L * 10000L) - 1L)`, respectively. Since the number of global quadrants in p4est is a signed int, I took this into account for the corresponding macro.

src/p4est_io.h
```diff
@@ -57,9 +57,12 @@ SC_EXTERN_C_BEGIN;
 #define P4EST_NUM_USER_STRING_BYTES 48 /**< number of user string bytes */
 #define P4EST_NUM_FIELD_HEADER_BYTES (2 + P4EST_NUM_ARRAY_METADATA_BYTES + P4EST_NUM_USER_STRING_BYTES)
                                      /**< number of bytes of one field header */
-#define P4EST_FILE_MAX_GLOBAL_QUAD 9999999999999999 /**< maximal number of global quadrants */
-#define P4EST_FILE_MAX_BLOCK_SIZE 9999999999999 /**< maximal number of block bytes */
-#define P4EST_FILE_MAX_FIELD_ENTRY_SIZE 9999999999999 /**< maximal number of bytes per field entry */
+#define P4EST_FILE_MAX_GLOBAL_QUAD ((10000L * 10000L * 10000L * 10000L) - 1L)
```
cburstedde, line 60: I'm still wondering. A factor of 10000UL is easy to confuse with 1000UL visually.

Shall we just revert this one and in the future define the maximum number of bytes, from which we calculate the maximum decimal number? That would be much easier on the eyes and brain.

tim-griesbach: I agree and revert the commit that introduces this product representation.

cburstedde:
Ok anything left for now?

tim-griesbach:
No, I do not think that there is anything left for now.

PR #154, opened 2022-09-23
Feature darwin ci

gghosh8495:
Added new ci_darwin.yml file to test p4est builds on MacOS.


cburstedde:
Thanks; just testing a couple tweaks before pushing and mergin.

PR #155, opened 2022-09-26
Simulation restart example

tim-griesbach:
# Simulation restart example

Proposed changes:
- Implement a simulation restart example for the existing p4est example `p{4,8}est_step3`.
- The behavior of `step3` stays unchanged if the checkpointing or loading of a checkpoint file is not invoked by command line parameters.
- This example checks the endianness of the read checkpoint file but the example just aborts if the endianness is wrong.

This example is a further demonstration of the functionalities implemented in https://github.com/cburstedde/p4est/pull/149 and https://github.com/cburstedde/p4est/pull/153.


cburstedde:
Thanks! Please study which arguments must be passed to `sc_options_add_*`. The long argument is constructed from the option name and must be suitable for the command line, to produce something like `--write`, but not be a sentence with spaces or other funny characters. Dashes are ok if sensible: the name `option-foo` becomes `--option-foo`.

tim-griesbach:
> Thanks! Please study which arguments must be passed to `sc_options_add_*`. The long argument is constructed from the option name and must be suitable for the command line, to produce something like `--write`, but not be a sentence with spaces or other funny characters. Dashes are ok if sensible: the name `option-foo` becomes `--option-foo`.

I fixed the long strings for the command line parameters.

example/steps/p4est_step3.c
```diff
@@ -1489,9 +1489,9 @@ main (int argc, char **argv)
 
   /* read command line options */
   opt = sc_options_new (argv[0]);
-  sc_options_add_bool (opt, 'C', "write checkpoints", &step3_checkpoint, 0,
+  sc_options_add_bool (opt, 'C', "write-checkpoints", &step3_checkpoint, 0,
```
cburstedde, line 1492: Singula please. :)

tim-griesbach: Right, we only write once at the end of the simulation.

PR #157, opened 2022-09-29
Topdown ghost interface

gghosh8495:
Initial iteration for the new ghost interface 

My opinion is that we create 'p4est_ghost_ext.h' where we move all functions and structs which are common to both 'p4est_ghost_ext.h' and 'p4est_ghost.h'. Then we #include p4est_ghost_ext.h into p4est_ghost.c, this way we avoid all code duplication.

The user can use a single header to use all relevant ghost stuff.

@cburstedde What would be your opinion ? You can refer to 'p4est_ghost_ext.h', you can skip the source files for now.


gghosh8495:
@cburstedde Ignore the CI failure, that is because p8est files are still under construction.

PR #158, opened 2022-09-30
Fix string truncation warnings

tim-griesbach:
# Fix string truncation warnings

Proposed changes: This PR avoids warnings due to string truncation by the usage of `strncpy`. These warnings are prevented by using the dedicated function `sc_strcopy` (cf. https://github.com/cburstedde/libsc/pull/89) in libsc.


cburstedde:
Thanks, do we need this merge commit? I just pushed an updated prev3-develop.

tim-griesbach:
I do not think that we need this merge commit. I can force push this branch to the single commit if you want.

PR #159, opened 2022-09-30
Feature cmake valgrind

gghosh8495:


gghosh8495:
@cburstedde By default CMake in p4est creates a static library so I don't think anything extra needs to be done.

cburstedde:
Thanks; I'm getting on the CI page:

No files were found with the provided path: ./Testing/Temporary/. No artifacts will be uploaded

gghosh8495:
I observed that when the memcheck passes the files are empty. Maybe the CI does not do upload blank chunks of memory.

cburstedde:
> I observed that when the memcheck passes the files are empty. Maybe the CI does not upload blank chunks of memory. I believe data is written to files only when leaks are detected.

Thanks, could you please verify this  supposition?

gghosh8495:
@cburstedde The memcheck logs are now uploaded correctly. The path needed to be changed.

cburstedde:
Thanks; I've used this PR to integrate a couple small updates.

PR #161, opened 2022-10-23
Introduce write/read for p4est and connectivity

tim-griesbach:
# Introduce write/read for p4est and connectivity

Proposed changes:
- Introduce `p{4,8}est_file_{write,read}`  to write and read a p4est.
- Introduce `p{4,8}est_file_{write,read}_connectivity`.
- Use the new functions in the simulation restart code in `p4est_step3`.
- Write a checking number instead of a bool in `p4est_step3` to check the endianness.
- Return `NULL` in `p4est_connectivity_inflate` in case of error.


src/p4est_io.c
```diff
@@ -1796,6 +1796,9 @@ p4est_file_error_string (int errclass, char *string, int *resultlen)
   case P4EST_FILE_ERR_SECTION_TYPE:
     tstr = "Valid non-matching section type";
     break;
+  case P4EST_FILE_ERR_CONN:
+    tstr = "Invalid serialized connectivty data";
```
cburstedde, line 1800: Typo

cburstedde:
Thanks; a couple thoughts:

 - check behavior of p4est_connectivity_inflate on erroneous data
 - document the user string size for the new read and write functions
 - rename to p4est_file_write_p4est etc.

src/p4est_io.c
```diff
@@ -2029,7 +2031,8 @@ p4est_file_write_connectivity (p4est_file_context_t * fc,
     p4est_file_write_block (fc, sizeof (size_t),
                             &conn_size_arr,
                             P4EST_STRING " connectivity size", errcode);
-  if (fc == NULL || *errcode != P4EST_FILE_ERR_SUCCESS) {
+  if (*errcode != P4EST_FILE_ERR_SUCCESS) {
+    P4EST_ASSERT (fc == NULL);
     sc_array_destroy (conn_buffer);
     return NULL;
   }
```
cburstedde, line 2038: Would it be even more strict to generally add?

    else { P4EST_ASSERT (fc != NULL); }

src/p4est_io.c
tim-griesbach: Right, I fixed the typo.

src/p4est_io.c
tim-griesbach: Yes, I can add this. However, in the relevant code places fc is passed to a p4est_file* function if the error code was P4EST_FILE_ERR_SUCCESS and these functions already assert fc != NULL. Therefore, I think it would be a doubled assertion.

cburstedde: Ok, leaving it sounds reasonable.

PR #163, opened 2022-11-17
Check read p4est data

tim-griesbach:
# Check read p4est data

Proposed changes:
- Introduce a new error code for invalid p4est data.
- The `pertree` array is checked for validity after reading.
- Introduce the new function `p4est_inflate_null` that frees the p4est in case of invalidity and returns `NULL`.
- Introduce `p4est_coordinates_{is_valid,is_inside_root}` to check the read quadrant coordinates.
- Read and write compressed quadrants as arrays and not as structs. In particular, this avoids problems due to compiler padding.
- More explicit documentation of the case of an invalid connectivity.
- Complete memory clean up in case of an error.
- Add a missing reshape call.


src/p4est_bits.c
```diff
@@ -657,17 +657,31 @@ p4est_quadrant_child_id (const p4est_quadrant_t * q)
...
 
 int
-p4est_quadrant_is_inside_root (const p4est_quadrant_t * q)
+p4est_coordinates_is_inside_root (const p4est_qcoord_t coord[])
 {
-  return
-    (q->x >= 0 && q->x < P4EST_ROOT_LEN) &&
-    (q->y >= 0 && q->y < P4EST_ROOT_LEN) &&
+  return (coord[0] >= 0 && coord[0] < P4EST_ROOT_LEN) && (coord[1] >= 0
+                                                          && coord[1] <
```
cburstedde, line 663: Can this be wrapped a bit more nicely?

src/p4est_io.c
```diff
@@ -36,6 +36,9 @@
 #include <sc_search.h>
 #include <sc.h>
 
+#define P4EST_FILE_COMPRESSED_QUAD_SIZE ((P4EST_DIM + 1) * sizeof (p4est_qcoord_t)) /**< size of a compressed
```
cburstedde, line 39: Please break into multiple lines.

src/p4est_io.c
```diff
@@ -2043,12 +2077,38 @@ p4est_file_read_p4est (p4est_file_context_t * fc, p4est_connectivity_t * conn,
...
+    *errcode = P4EST_FILE_ERR_P4EST;
+    sc_array_reset (&pertree_arr);
+    P4EST_FILE_CHECK_NULL (*errcode, fc,
+                           P4EST_STRING "_file_read_" P4EST_STRING, errcode);
+  }
+  for (jt = 0; jt < conn->num_trees; ++jt) {
+    if (!(pertree[jt] <= pertree[jt + 1])) {
+      *errcode = P4EST_FILE_ERR_P4EST;
+      sc_array_reset (&pertree_arr);
+      P4EST_FILE_CHECK_NULL (*errcode, fc,
```
cburstedde, line 2092: We jump out of a loop by goto, this seems fine (?).

cburstedde:
Thanks, looks very nice!

src/p4est_io.c
```diff
@@ -2043,12 +2077,38 @@ p4est_file_read_p4est (p4est_file_context_t * fc, p4est_connectivity_t * conn,
...
+                           P4EST_STRING "_file_read_" P4EST_STRING, errcode);
+  }
+
   gfq = P4EST_ALLOC (p4est_gloidx_t, mpisize + 1);
   /** Compute a uniform global first quadrant array to use a uniform
    * partition to read the data fields in parallel.
    */
   p4est_comm_global_first_quadrant (fc->global_num_quadrants, mpisize, gfq);
 
+  P4EST_ASSERT (gfq[mpisize] == pertree[conn->num_trees]);
```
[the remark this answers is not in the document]
tim-griesbach: Yes, we check `pertree[conn->num_trees]` in line 2097 above and `gfq` is calculated with `fc->global_num_quadrants` what is also used to check  `pertree[conn->num_trees]`. Therefore, we are certain about this statement independent of what the user does.

src/p4est_io.c
tim-griesbach: This macro `P4EST_FILE_CHECK_NULL` is not a an actual goto but a clean up and error code processing followed by a `return NULL`.
Therefore, I think this clean up, error code processing and return in the loop should be fine.

The error macros only set goto if they are dedicated to be called on only one rank since then we need to synchronize at the first collective code place.

src/p4est_io.c
tim-griesbach: I pushed an according change.

PR #164, opened 2022-11-24
Added new ghost files to CMake build

gghosh8495:


gghosh8495:
@cburstedde This should fix the cmake build.

PR #167, opened 2022-12-15
Feature tree boundary

gghosh8495:
Proposal for a function to check if two related quadrants share a face.


gghosh8495:
@cburstedde The CMake build error originates from apt-get package manager mismatch. Tim has put up a PR that fixes that issue. Please, merge Tim's PR before merging this to avoid redundant CI failure.

gghosh8495:
@cburstedde Yes, the CI error goes away.

PR #169, opened 2022-12-20
Function to check for shared corners

gghosh8495:
Proposal for a function to check if two related quadrants have common corners.


cburstedde:
Thanks; please use the optimized calculation we talked about offilne.

cburstedde:
With our recent merges, there are now conflicts probably due to white space at the end of the file. In the updated version coming up, it should be possible to provide a conflict-free diff.

cburstedde:
Thanks! Looks nice.

Just a style/convention comment: in the .h files, we usually have the face functions before the corner functions. We also have one empty line between the end of one prototype and the documentation of the next one. Would you be able to make these modifications?

gghosh8495:
@cburstedde Would you like to have any more changes ?

cburstedde:
Thanks, looks good, I just made one tiny swap on top of it.

Issue #178, opened 2023-02-11
Handle multiple libraries interacting with p4est/sc


sloede:
**Description**
If I understand correctly, `sc_init`/`p4est_init` and `sc_finalize` are only supposed to be called once *per process*. This makes it hard to use p4est in a setting where multiple libraries are linked in an executable which individually use p4est, since as far as I can tell, there is no official way to check if sc/p4est have already been initialized/finalized. Thus, each library that uses p4est has either to assume that it is the only library to do so and call `p4est_init` unconditionally, or they have to provide for the possibility that it is initialized elsewhere and manually disable the call to `p4est_init`.

For easier composability of libraries, it would thus be great if sc/p4est were to provide an API that provides this information and allows one to safely compose multiple libraries that make use of p4est.

**Proposed solution**
Following the MPI standard and its API for [`MPI_Initialized`](https://www.open-mpi.org/doc/v4.1/man3/MPI_Initialized.3.php) and [`MPI_Finalized`](https://www.open-mpi.org/doc/v4.1/man3/MPI_Finalized.3.php), one possible solution would be if sc/p4est were to provide the following public API,
```c
void p4est_initialized(int* flag);
void sc_initialized(int* flag);
void sc_finalized(int* flag);
```
where one can call each function with a pointer to an `int` that will either be `0` or `1` afterwards, depending on whether sc/p4est have already been initialized. This would allow to guard the init/finalization code in libraries with something like
```c
int initialized;
sc_initialized(&initialized);
if (!initialized) {
  sc_init(MPI_COMM_WORLD, 0, 0, NULL, SC_LP_ERROR);
}
p4est_initialized(&initialized);
if (!initialized) {
  p4est_init(NULL, SC_LP_ERROR);
}

[...]

int finalized;
sc_finalized(&finalized);
if (!finalized) {
  sc_finalize();
}
```

If API similarity to MPI is not relevant/desirable, alternative APIs could be
* a function like `p4est_is_initialized`, which directly returns true or false, or
* a function like `p4est_init_idempotent`, which internally checks for initialization and does nothing if already initialized.

cc @ranocha @kronbichler @lchristm @JoshuaLampert

cburstedde:
The init and finalize functions are somewhat soft in that they are not strictly required. However, they do set the internal rank variable, which is used in logging output, and then they catch some signals if so desired. I agree this should be done once per program and not once per sub-application.

I like the is_initialized and is_finalized method the best. Idempotent versions may be written by any user.

[...]

PR #180, opened 2023-03-03
Add function to check if p4est has been initialized

sloede:
This adds the first part of the functionality as discussed in #178, i.e., the ability to check if p4est has already been initialized through a function `p4est_is_initialized`.

I'm not sure if this needs anything else - or if this is even the right target branch for the next release - but I just thought I'd get the discussion started by opening a PR to iterate on.

cc @ranocha

### DIFF
--- a/src/p4est_base.c
+++ b/src/p4est_base.c
@@ -25,7 +25,7 @@
 #include <p4est_base.h>
 
 int                 p4est_package_id = -1;
-int                 p4est_is_initialized_ = 0;
+int                 p4est_initialized = 0;
 
 void
 p4est_init (sc_log_handler_t log_handler, int log_threshold)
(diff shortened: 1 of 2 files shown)


cburstedde:
Thanks; would you like to un-initialize on sc_finalize, such that we can have an arbitrary number of p4est life cycles?

cburstedde:
(We are not using trailing underscore variables_ in p4est, even though it's a cool concept.)

sloede:
> (We are not using trailing underscore variables_ in p4est, even though it's a cool concept.)

Oh, I can change that back if you want. I just thought it would be good to indicate that this is not a publicly available field, since it is not declared `extern`.

cburstedde:
> > Thanks; would you like to un-initialize on sc_finalize, such that we can have an arbitrary number of p4est life cycles?
> 
> I am not sure if this is really needed though. You can already create a (somewhat) arbitrary number of meshes, can't you? I don't see a big benefit in allow to initialize/de-initialize multiple times. Thus if it is up to me, I'd propose to proceed with this and allow de-initialization to be tackled in a future PR.
> 
> My main motivation for this PR was that we want to couple two libraries that both use p4est and make sure that only one of them initializes p4est. Right now we check for the `p4est_package_id` to be >=0, but I thought a proper interface would be nicer.

You have a point, of course, but technically not resetting is_initialized to
0 in calling sc_finalize is an error.  These functions are designed to be
inverses of each other, and the status variable should please reflect that.

sloede:
> You have a point, of course, but technically not resetting is_initialized to 0 in calling sc_finalize is an error. These functions are designed to be inverses of each other, and the status variable should please reflect that.

Ah, now I understand what you mean. However, I am not sure how to do that: As far as I can tell, libsc is unaware of the package from which it is called, so how could I reset a variable (or even call a function) in p4est from `sc_finalize`?

cburstedde:
About to merge after a non-functional doc update. Thanks!

The whole init/finalize logic, also with libsc, would need some thinking.
It would be cool to sit down and talk about the needs from the wrapping perspective.

PR #181, opened 2023-03-17
Deprecate data file format

tim-griesbach:
# Deprecate data file format

Proposed changes: Since we are planning to release an updated parallel data file format for p4est, we deprecate the old one to clarify that it will be replaced soon. The old data file format and all corresponding functionality like the simulation restart in `p{4,8}est_step3` can be activated by `--enable-file-deprecated` as configure flag for autotools or via the variable `enable-file-deprecated` in CMake.

The `p{4,8}est_{load,save}` functionality is unaffected and besides the functions for the data file format `p{4,8}est_file_*` including its tests and the restart functionality in `p{4,8}est_step3` nothing is changed and in particular no other existing functionality of p4est breaks.


cburstedde:
Thanks! I have made some minimal changes, also (against our original idea to avoid reindintation) removed the goto.

If this works, please feel free to merge yourself.

PR #190, opened 2023-03-31
CI:CMake: add GCC-12, work with CMake 3.26 runner default

scivision:
Just like https://github.com/cburstedde/libsc/pull/113

For GitHub Actions CI:

* GCC 12 is now also available
* CMake 3.26 is now used, which does not have CMake{Error,Output}.log anymore, replaced by CMakeConfigureLog.yaml
* CTEST_NO_TESTS_ACTION: "error" is in case something is amiss in the CMakeLists.txt that accidentally disabled all tests, this environment variable fails CTest if no tests are enabled.

As I had done previously in libsc CMake CI, to make the CI tests run faster, I only test with GCC 9-12

### DIFF
--- a/.github/workflows/ci_cmake.yml
+++ b/.github/workflows/ci_cmake.yml
@@ -89,8 +89,7 @@ jobs:
       with:
         name: linux_cmake_log
         path: |
-          ./build/CMakeFiles/CMakeOutput.log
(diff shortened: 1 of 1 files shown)


cburstedde:
Thanks!

cburstedde:
I am getting a CI failure now: The workflow script appears invalid. Can you please check?

PR #193, opened 2023-04-03
cmake:example: don't try to run test in CI

scivision:
The examples were not intended to be run in the CI


cburstedde:
Thanks a lot!

PR #196, opened 2023-04-12
Introduce p4est_search_partition_gfp

hannesbrandt:
We add p4est_search_partition_gfp as an alternative to p4est_search_partition_gfx.
The parameter gfq is no longer required as a parameter, which may avoid unnecessary communication in case gfq is not already available locally, but makes the function slightly slower.
To implement p4est_search_partition_gfp with minimal changes, we add p4est_comm_is_empty_gfx, which works like p4est_comm_is_empty_gfq for gfq != NULL and else calls p4est_quadrant_is_equal_piggy on gfp.

### DIFF
--- a/src/p4est_communication.c
+++ b/src/p4est_communication.c
@@ -596,6 +596,21 @@ p4est_comm_is_empty_gfq (const p4est_gloidx_t *gfq, int num_procs, int p)
 }
 
 int
+p4est_comm_is_empty_gfx (const p4est_gloidx_t *gfq, const p4est_quadrant_t *gfp,
+                         int num_procs, int p)
+{
+  P4EST_ASSERT (0 <= p && p < num_procs);
+  P4EST_ASSERT (gfp != NULL);
+
+  if (gfq) {
+    return gfq[p] == gfq[p + 1];
+  }
+  else {
+    return p4est_quadrant_is_equal_piggy (&gfp[p], &gfp[p + 1]);
+  }
+}
+
+int
 p4est_comm_is_contained (p4est_t * p4est, p4est_locidx_t which_tree,
                          const p4est_quadrant_t * q, int rank)
 {
--- a/src/p8est_search.h
+++ b/src/p8est_search.h
@@ -454,6 +454,43 @@ void                p8est_search_partition_gfx
    p8est_search_partition_t quadrant_fn, p8est_search_partition_t point_fn,
    sc_array_t *points);
 
+/** Traverse some given global partition top-down.
+ * The partition can be that of any p4est, not necessarily known to the
+ * caller.  This is not a collective function.  It does not communicate.
+ * We proceed top-down through the partition, identically on all processors
+ * except for the results of two user-provided callbacks.  The recursion will only
+ * go down branches that are split between multiple processors.  The callback
+ * functions can be used to stop a branch recursion even for split branches.
+ * This function offers the option to search for arbitrary user-defined points
+ * analogously to \ref p4est_search_local.
+ * This function is similar to p4est_search_partition_gfx, but does not require
+ * the p4est_gloidx_t array gfq. If gfq is available, using
+ * p4est_search_partition_gfx is recommended, because it is slightly faster.
+ * \note Traversing the whole given partition will be at least O(P),
+ *       so sensible use of the callback function is advised to cut it short.
+ * \param [in] gfp          Partition position to traverse.  Length \a nmemb + 1.
+ * \param [in] nmemb        Number of processors encoded in \a gfp (plus one).
+ * \param [in] num_trees    Tree number must match the contents of \a gfp.
+ * \param [in] call_post    If true, call quadrant callback both pre and post
+ *                          point callback, in both cases before recursion (!).
+ * \param [in] user         We pass a dummy p4est to the callbacks whose only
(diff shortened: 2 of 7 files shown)


src/p4est_communication.c
```diff
@@ -595,6 +595,21 @@ p4est_comm_is_empty_gfq (const p4est_gloidx_t *gfq, int num_procs, int p)
   return gfq[p] == gfq[p + 1];
 }
 
+int
+p4est_comm_is_empty_gfx (const p4est_gloidx_t *gfq, const p4est_quadrant_t *gfp,
+                         int num_procs, int p)
+{
+  P4EST_ASSERT (0 <= p && p < num_procs);
+  P4EST_ASSERT (gfp != NULL);
```
cburstedde, line 603: Minor detail, please check arguments in order of the prototype (gfp before num_procs, p)

src/p4est_communication.c
```diff
@@ -595,6 +595,21 @@ p4est_comm_is_empty_gfq (const p4est_gloidx_t *gfq, int num_procs, int p)
...
+
+  if (gfq) {
```
cburstedde, line 605: Please check pointers explicitly with == NULL or  != NULL.

src/p8est_search.h
```diff
@@ -454,6 +454,43 @@ void                p8est_search_partition_gfx
...
+/** Traverse some given global partition top-down.
+ * The partition can be that of any p4est, not necessarily known to the
+ * caller.  This is not a collective function.  It does not communicate.
+ * We proceed top-down through the partition, identically on all processors
+ * except for the results of two user-provided callbacks.  The recursion will only
+ * go down branches that are split between multiple processors.  The callback
+ * functions can be used to stop a branch recursion even for split branches.
+ * This function offers the option to search for arbitrary user-defined points
+ * analogously to \ref p4est_search_local.
+ * This function is similar to p4est_search_partition_gfx, but does not require
```
cburstedde, line 466: You may use the \ref doxygen command to generously cross-link functions. I would use it wherever a linkable prototype occurs. Please check the generated html for consistent looks and remove warnings as you go.

cburstedde:
> Thanks! Strictly speaking, if gfq is optional in the comm _gfx function, should it be optional in the partition _gfx function, too? So the assertions only execute if not NULL.

In that case partition_gfp might directly call partition_gfx (NULL, p).

hannesbrandt:
I changed the assertions in `p4est_search_partition_gfx` to check for `gfq == NULL || the previous assertions`. Now `p4est_search_partition_gfp` calls `p4est_search_partition_gfx` with `NULL` for the `gfq` parameter.
I added some doxygen commands and made some minor changes in the `search.h` files like p4est functions being referenced in the p8est file or references not being recognized due to a trailing `:`.

cburstedde:
Thanks for the update!

I'm thinking about a more philosophical point: is it a good thing if some arguments of a function may or may not be NULL, legally. This increases coding flexibility but places more burdens on the user to read the API docs in detail. If they do not read them carefully, chances are high that they do something suboptimal. If we were to not allow for NULL defaults wherever sensible, the user might be a happier person and the code might be more stable.

So I know I suggested that the gfq argument is always optional for the gfx functions.
The alternative would be to remove is_empty_gfx, leaving the _gfp and _gfq versions, and to require gfq to always be non-NULL in search_partition_gfx (if I'm not confusing things). This way the user does not need to read the documentation about parameters that may or may not be NULL.

Whout would be your opinion, also checking with @tim-griesbach?

hannesbrandt:
That is a good point. For a more consistent setup @tim-griesbach  and I think we should
- reinstate the `qfg != NULL` check in `search_partition_gfx`
- change `search_partition_gfp` from calling `search_partition_gfx` to calling `search_partition_internal`
- add `is_empty_gfp` to p4est_communication
- remove `is_empty_gfx` and replace every call of this function in `search_partition_internal `by something like
  ```(gfq != NULL && is_empty_gfq)  || (gfq == NULL && is_empty_gfp)```

What do you think?

cburstedde:
If you might briefly comment on my conversations above I'll be happy to resolve them.

src/p8est_search.h
hannesbrandt: Good point, I added several references to the documentation.

src/p4est_communication.c
hannesbrandt: Good point, I've changed this to `gfq != NULL` now.

hannesbrandt:
>     * reinstate the `qfg != NULL` check in `search_partition_gfx`
> 
>     * change `search_partition_gfp` from calling `search_partition_gfx` to calling `search_partition_internal`
> 
>     * add `is_empty_gfp` to p4est_communication
> 
>     * remove `is_empty_gfx` and replace every call of this function in `search_partition_internal `by something like
>       `(gfq != NULL) ? is_empty_gfq : is_empty_gfp`

I've implemented everything according to the list.
Now, the only `_gfx` function remaining is `search_partition_gfx`, which actually uses both `gfx` and `gfp`. All other related functions are named either `_gfq` or `_gfp` and always assume that the respective parameter is not `NULL`. I've used a macro in `search.c` to distinguish the cases in `search_partition_internal` like you proposed.

PR #200, opened 2023-04-19
Add brick connectivity to simple example

pkestene:
# Add brick connectivity to simple example

Very minor changes, just slightly update simple examples to demo the "brick" connectivity (illustrate tree numbering).


cburstedde:
Thanks again! Useful to display the brick, indeed.

Issue #211, opened 2023-04-30
points example lacking point data


pkestene:
**Description**

Currently the "points" examples are disabled in the cmake build. 
These examples require pts data files (one file per MPI process).

There is also a comment in example/CMakelist.txt stating the points data files are missing, so that why the example is disabled in the cmake build (while still enabled in the autotools build) .

**Proposed solution**

- provide a simple executable that generate some point data to avoid distributing binary data files
- instead of generating one point data file per MPI process, just use MPI-IO to make reading / writing the point data file a MPI collective operation. One advantage is that user can easily generate data for any number of MPI processes, and read them with a possibly different number of MPI processes as easily.

cburstedde:
Thanks for thinking this forward!

cburstedde:
I'm merging #212, which resolves the issue. We may keep this issue open low-priority to play with the location of the generated points some more depending on the connectivity.

Likewise, adding the sc_io error handling will be instructive with this example.

cburstedde:
While we do not distribute a points data file, we have a program to generate points on demand. Currently, the points are created randomly within a cube. This works well if the domain to generate is indeed the unit cube. For other connectivities/geometries, it would be desirable to query the bounding box of each tree and to generate points within their domains. Alternatively, we may add options to generate points according to an analytic distribution with a support wide enough that it covers all of our predefined geometries.

PR #212, opened 2023-04-30
Update points example

pkestene:
# Update points example

Following up on issue #211

Proposed changes:
- provide a helper executable to generate example point data
- alos add the brick connectivity (with a slightly different treatment, just for illustrative purpose)
- use MPI-IO to create the point data file (a single file for all MPI processes).

Currently this can only be merged when companion MR https://github.com/cburstedde/libsc/pull/128 which update MPI-IO routines in libsc is merge to. 
The executable that generate point data file need these mpi-io routines in libsc.

### DIFF
--- a/example/points/Makefile.am
+++ b/example/points/Makefile.am
@@ -3,8 +3,6 @@
 # Makefile.am in example/points
 # included non-recursively from toplevel directory
 
-#if P4EST_ENABLE_MPIIO
-
 if P4EST_ENABLE_BUILD_2D
 bin_PROGRAMS += example/points/p4est_points
 example_points_p4est_points_SOURCES = example/points/points2.c
@@ -24,5 +22,3 @@ example_points_p8est_points_generate_SOURCES = example/points/generate_points3.c
 
 LINT_CSOURCES += $(example_points_p8est_points_SOURCES)
 endif
-
-#endif
--- a/example/points/Makefile.am
+++ b/example/points/Makefile.am
@@ -3,12 +3,14 @@
 # Makefile.am in example/points
 # included non-recursively from toplevel directory
 
+if P4EST_ENABLE_MPIIO
+
 if P4EST_ENABLE_BUILD_2D
 bin_PROGRAMS += example/points/p4est_points
 example_points_p4est_points_SOURCES = example/points/points2.c
 
-bin_PROGRAMS += example/points/p4est_generate_points
-example_points_p4est_generate_points_SOURCES = example/points/generate_points.c
+bin_PROGRAMS += example/points/p4est_points_generate
+example_points_p4est_points_generate_SOURCES = example/points/generate_points2.c
 
 LINT_CSOURCES += $(example_points_p4est_points_SOURCES)
 endif
@@ -17,9 +19,10 @@ if P4EST_ENABLE_BUILD_3D
 bin_PROGRAMS += example/points/p8est_points
 example_points_p8est_points_SOURCES = example/points/points3.c
 
-bin_PROGRAMS += example/points/p8est_generate_points
-example_points_p8est_generate_points_SOURCES = example/points/generate_points.c
-example_points_p8est_generate_points_CPPFLAGS = -DTHREED $(AM_CPPFLAGS)
+bin_PROGRAMS += example/points/p8est_points_generate
+example_points_p8est_points_generate_SOURCES = example/points/generate_points3.c
 
 LINT_CSOURCES += $(example_points_p8est_points_SOURCES)
+endif
+
 endif
--- a/example/points/Makefile.am
+++ b/example/points/Makefile.am
@@ -3,7 +3,7 @@
 # Makefile.am in example/points
 # included non-recursively from toplevel directory
 
-if P4EST_ENABLE_MPIIO
+#if P4EST_ENABLE_MPIIO
 
 if P4EST_ENABLE_BUILD_2D
 bin_PROGRAMS += example/points/p4est_points
@@ -25,4 +25,4 @@ example_points_p8est_points_generate_SOURCES = example/points/generate_points3.c
 LINT_CSOURCES += $(example_points_p8est_points_SOURCES)
 endif
 
-endif
+#endif
--- a/example/points/Makefile.am
+++ b/example/points/Makefile.am
@@ -7,12 +7,19 @@ if P4EST_ENABLE_BUILD_2D
 bin_PROGRAMS += example/points/p4est_points
 example_points_p4est_points_SOURCES = example/points/points2.c
 
+bin_PROGRAMS += example/points/p4est_generate_points
+example_points_p4est_generate_points_SOURCES = example/points/generate_points.c
+
 LINT_CSOURCES += $(example_points_p4est_points_SOURCES)
 endif
 
 if P4EST_ENABLE_BUILD_3D
 bin_PROGRAMS += example/points/p8est_points
 example_points_p8est_points_SOURCES = example/points/points3.c
+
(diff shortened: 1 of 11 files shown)


example/points/Makefile.am
```diff
@@ -7,12 +7,19 @@ if P4EST_ENABLE_BUILD_2D
 bin_PROGRAMS += example/points/p4est_points
 example_points_p4est_points_SOURCES = example/points/points2.c
 
+bin_PROGRAMS += example/points/p4est_generate_points
```
cburstedde, line 10: Thanks; what about renaming to p4est_points_generate?

example/points/generate_points.c
```diff
@@ -0,0 +1,246 @@
...
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with p4est; if not, write to the Free Software Foundation, Inc.,
+  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+*/
+
+#ifdef THREED
```
cburstedde, line 25: This macro THREED is unusual in all of p4est. Users might think that this is a convention they should copy in their own code. Is it possible to rely on P4_TO_P8 exclusively, like in the rest of p4est?

example/points/Makefile.am
```diff
@@ -7,12 +7,19 @@ if P4EST_ENABLE_BUILD_2D
...
 LINT_CSOURCES += $(example_points_p4est_points_SOURCES)
 endif
 
 if P4EST_ENABLE_BUILD_3D
 bin_PROGRAMS += example/points/p8est_points
 example_points_p8est_points_SOURCES = example/points/points3.c
 
+bin_PROGRAMS += example/points/p8est_generate_points
+example_points_p8est_generate_points_SOURCES = example/points/generate_points.c
+example_points_p8est_generate_points_CPPFLAGS = -DTHREED $(AM_CPPFLAGS)
```
cburstedde, line 22: See below comment; such a THREED macro is usually not necessary in how we write dimension-dependent code.

example/points/generate_points.c
```diff
@@ -0,0 +1,246 @@
...
+#else
+#include <p4est_bits.h>
+#include <p4est_points.h>
+#include <p4est_vtk.h>
+#endif /* !P4_TO_P8 */
+#include <sc_io.h>
+
+
+static void
+generate_points (const char *filename,
```
cburstedde, line 42: Will this file compile when MPIIO is not enabled (try --disable-mpiio)? This relates to libsc, where MPI_File_get_view etc. are defined with MPIIO, but have no correspondence without it.

example/points/Makefile.am
pkestene: ok will do.

example/points/generate_points.c
pkestene: ok, I will turn `generate_points.c` into `generate_points2.c` and `generate_points3.c`and define P4_TO_P8 in the later.

example/points/generate_points.c
pkestene: ok, MPIIO is required. We can disable building this example when mpi-io is not available.

cburstedde: This seems fine for now. @tim-griesbach do you see a way to abstract the MPI I/O status away in the future?

cburstedde:
Thanks; I've merged the prerequisites in libsc!

I'm wondering whether the example itself may survive without MPIIO, just not the file generation -- is there a way?

example/points/generate_points.c
tim-griesbach: We can abstract the MPI I/O status away using the functions `sc_io_open`, `sc_io_write_at_all`, `sc_io_read_at_all` and `sc_io_close` since these function work also without MPI I/O or even without MPI.

However, in `points2.c` the number of points on reading is currently determined using the file size. Up to my understanding this does not work without MPI I/O in a general portable way by just relying on the C standard (cf. https://wiki.sei.cmu.edu/confluence/display/c/FIO19-C.+Do+not+use+fseek()+and+ftell()+to+compute+the+size+of+a+regular+file). 

We can solve this problem by writing the number of points to the file. Assuming this adjustment we shall obtain code that produces equivalent files with and without MPI I/O.

tim-griesbach: If you want, I can adjust the code as described above.

tim-griesbach:
> This sounds all good. If points works generally with a change to the file format and using `#ifdef P4EST_ENABLE_MPIIO` and/or sc_io wrappers, please go ahead!

I pushed changes to use the `sc_io` wrappers and to write/read the number of global points to/from the file.

During testing my changes I observed different behavior of `sc_io_open` in the create mode depending on the MPI IO status. These differences come from the difference between `fopen` with the mode `'w'` that truncates the file to length zero and `MPI_File_open` that just overwrites an existing file in the create mode.

Therefore, I adjusted `sc_io_open` such that it also truncates the file to length zero in the case the creation mode. The changes are in the branch https://github.com/tim-griesbach/libsc/tree/feature-io-improvements. I will create a PR for these libsc changes.

cburstedde:
Would it be possible to rebase this branch on the latest develop without touching sc in the history?

pkestene:
> Would it be possible to rebase this branch on the latest develop without touching sc in the history?

I think commit b571e91458ce6ae6ca686c29423c30efcf3320ab should be modified not to update sc git submodule; then the rebase would be easier.

cburstedde:
> > Would it be possible to rebase this branch on the latest develop without touching sc in the history?
> 
> I think commit b571e91458ce6ae6ca686c29423c30efcf3320ab should be modified not to update sc git submodule; then the rebase would be easier.

Is it possible to fix this during the rebase?  When basing on develop, the
sc submodule will always be the same at 04ec24f6869c08a7b76c14ce7684e9952.

pkestene:
Hi @cburstedde 
FYI, i've just rebase this onto latest develop and made simple adjustment to make it work again. I think it is in good shape now.

cburstedde:
> Hi @cburstedde FYI, i've just rebase this onto latest develop and made simple adjustment to make it work again. I think it is in good shape now.

Thanks so much, I'll take a look!

cburstedde:
Thanks! This is all very nice now. You gave me some motivation to write some outstanding documentation as well, I hope you don't mind I pushed it onto your PR. I've also used the p4est partition formulas for reference, even though your formulas work just fine as well!

I see that the generation of points is independent of the connectivity. I'm wondering whether we should do something about that, but otoh we may all have better things to do right now. I'm ready to merge as is.

cburstedde:
A note to @tim-griesbach: This example nicely demonstrates use of some sc_io functions. How hard would it be to introduce proper error handling, to make this example instructive and representative of safe I/O code in p4est?

tim-griesbach:
> A note to @tim-griesbach: This example nicely demonstrates use of some sc_io functions. How hard would it be to introduce proper error handling, to make this example instructive and representative of safe I/O code in p4est?

It should be easy to add in the case of an I/O error the conversion of the error code to an error string using `sc_MPI_Error_class`. This works independent of the availability of MPI (I/O) since the `sc_io` functions internally translate the error codes to MPI error classes or wrappers of these MPI error classes that work without MPI (I/O) and `sc_MPI_Error_string` also works without MPI I/O. 

Then we could print the error string to inform the user about the specific error. If an I/O error occurs we would still abort after printing the error string.

cburstedde:
Thanks all, I've updated #211 with the current status and ammerging now.

Issue #213, opened 2023-05-04
cmake variable sc_external used before defined


pkestene:
**Description**

In top-level CMakeLists.txt variable `sc_external` is used at line 33 (https://github.com/cburstedde/p4est/blob/master/CMakeLists.txt#L33), while it is defined inside `cmake/sc.cmake` that is not sourced yet.

**Proposed solution**

As sc_external behave like an option, just make it actually a cmake option and move its definition inside `cmake/options.cmake`

cburstedde:
Thanks; this appears resolved! :)

PR #214, opened 2023-05-04
cmake variable sc_external is now a cmake option.

pkestene:
addressing issue #213


cburstedde:
Hi @scivision we have made several CMake updates for flexibility and alignment with autotools.
Are you on board with the latest status/any comments?

Issue #227, opened 2023-08-15
Build fail  without jansson


mkirilin:
**Description**
Build with cmkae fails on a system without jansson library.

**To Reproduce**
1. p4est repository is on  2e44 commit. sc repository is on 83bd.
2. Configure and build sc in debug mode and with mpi.
3. Configure and build p4est in debug and with mpi.

p4est configure output: `-- Could NOT find jansson (missing: jansson_DIR)`
p4est build output: 

```
/opt/software/easybuild/software/binutils/2.38-GCCcore-11.3.0/bin/ld: ../local/lib/libsc.a(sc_options.c.o): in function `sc_options_json_lookup':
sc_options.c:(.text+0x95): undefined reference to `json_object_get'
/opt/software/easybuild/software/binutils/2.38-GCCcore-11.3.0/bin/ld: sc_options.c:(.text+0x105): undefined reference to `json_object_get'
```
etc...

pkestene:
Hi @mkirilin 

Could you provide the full commit hash (at least first 7 characters) because I was not able to identify 2e44 in p4est repository, while I found 83bd019 in libsc (current master) ?

I tried to build both current p4est develop branch and master branch on a system where libjansson is not available; I didn't encounter any problem, it builds fine.

mkirilin:
Hi @pkestene 

the relevant commit is on eb24772a.
Some more details: compiler version is GCC 11.3.0. CMake version is 3.23.1 compiled with GCC 11.3.0.

pkestene:
@mkirilin 
Please re-check on your side.

eb24772 is current master of p4est.
This commit builds fine for me.

Make sure that libsc submodule is up-to-date inside your p4est clone (it should be on master too).

mkirilin:
@pkestene 
I double checked current commit states: both p4est and sc are on master.

The problem appears when jansson is not found via find_package() but is found via PkgConfig:
```
-- Could NOT find jansson (missing: jansson_DIR)
-- Found PkgConfig: /usr/bin/pkg-config (found version "0.29.2") 
-- jansson library found via pkg-config
```

pkestene:
@mkirilin 

Ok. I understand now and I can reproduce. This is a bug.
I'll push a fix shortly.

pkestene:
@mkirilin 

you can try this branch : https://github.com/pkestene/p4est/tree/fix/jansson-pkgconfig for cross-checking

mkirilin:
@pkestene 
[...]

PR #230, opened 2023-08-24
Fix linking to lib jansson (when available).

pkestene:
This is a follow-up; this issue was not entirely fixed in https://github.com/cburstedde/p4est/pull/229

Following up on issue #227


mkirilin:
Thanks for the fix. It builds well now on my machine.

PR #232, opened 2023-08-25
Doxygen: Fix warnings in p4est_connectivity

SAutum:
Doxygen: deal with warnings in p4est_connectivity.h during document generation.

# Warnings in p4est_connectivity.h during document generation


cburstedde:
Thanks, looking good so far! Please edit this PR text to format titles vs. text. For example "Title" is just a template that can be changed to the actual (short) title of the PR.

PR #234, opened 2023-08-30
p4est_search: fix to pass true quadrant

cburstedde:
In the relatively new function p4est_search_reorder, we accept a parameter to switch level skipping on or off as an optimization. If on and only a single leaf is contained in a search quadrant, we directly skip down to the leaf as next quadrant to consider.

We fix the bug that the quadrant was not real when skipping was off.


cburstedde:
Any other comments on search_reorder?

hannesbrandt:
This looks good to @tim-griesbach and me!

PR #235, opened 2023-08-30
add test_mesh_corners3.c

scottaiton:
Here's the test program that @hannesbrandt requested, which demonstrates the missing corner information I'm running into with p8est_mesh.


cburstedde:
Thanks! Can I ask you to add a file doc/author_aiton.txt like the others to release your code under BSD? We'll take care updating the Makefiles then.

hannesbrandt:
We discuss potential changes to the `p8est_mesh_t` regarding the edge-hanging corner neighbors in https://github.com/cburstedde/p4est/issues/244.

cburstedde:
Thanks, if I may just ask to format the test program according to our customs (e. g. indentation, no declarations after code within a block) and slightly beyond (liberally adding comments every couple lines explaining what's going on).

hannesbrandt:
Yes, I'll work on this.
During my work on PR https://github.com/cburstedde/p4est/pull/262 I also implemented an inter-tree test case that checks the implementation of edgehanging corner neighbors across tree faces and edges. I would propose to add this to the PR as well.

cburstedde:
> Yes, I'll work on this.
> During my work on PR https://github.com/cburstedde/p4est/pull/262 I also implemented an inter-tree test case that checks the implementation of edgehanging corner neighbors across tree faces and edges. I would propose to add this to the PR as well.

Sounds perfect!

cburstedde:
Thanks! Sorry for being obnoxious, but I'd caution against using TRUE variables in general. It is too tempting to compare == TRUE, which would be wrong. I'm fine with assigning = FALSE or = 0  or = !FALSE or = 1, which is legal.

cburstedde:
Nice thanks so much!

PR #236, opened 2023-08-31
Example: simple

SAutum:
# Three new examples related to p4est_connectivity

Proposed changes:
- Add three new examples (circle, drop and bowtie) into simple example.
- fix the warnings while generating documents in p4est_connectivity.
- Complete a navigation page of examples for users to get familiar with the app.
- *Add Hongli's contribution agreement under FreeBSD license.


cburstedde:
Thanks again, looks nice!

PR #239, opened 2023-09-25
Feature cmake example

mkirilin:
# Add examples into build by default

This PR adds examples into p4est build by default, namely:
1. The directories structure stays the same as in sources
2. Resource files are copied into build directories when it is necessary
3. Examples are now excluded from ctest
4. CI is edit to exclude redundant work related to installing libraries necessary for examples


example/CMakeLists.txt
```diff
@@ -1,35 +1,8 @@
-cmake_minimum_required(VERSION 3.13...3.24)
-project(p4estExamples LANGUAGES C)
-
-include(CheckSymbolExists)
-include(CheckIncludeFile)
-
 option(mpi "use MPI" off)
```
pkestene, line 1: you should probably remove this line: `mpi` option is already defined at top-level in `cmake/options.cmake`; it was fine when examples were in separate cmake project, but not anymore here.

example/CMakeLists.txt
```diff
@@ -1,35 +1,8 @@
...
 
 # --- find external libraries
 list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/../cmake)
```
pkestene: you probably do need this line anymore; same cmake project, CMAKE_MODULE_PATH should be ok, not need to modify it

example/CMakeLists.txt
mkirilin: Thanks for your feedback. The option `mpi` was redundant and the building works fine without it.

example/CMakeLists.txt
mkirilin: Thanks. This line is removed at the latest state if the branch.

cburstedde:
Thanks to @mkirilin for the simplification and @pkestene for the review. Does everything work as expected?

pkestene:
> Thanks to @mkirilin for the simplification and @pkestene for the review. Does everything work as expected?

LGTM

PR #240, opened 2023-09-26
add heister as author

tjhei:
add myself


cburstedde:
Thanks so much @tjhei !

Issue #238, opened 2023-09-13
3d connectivity balancing issue


SAutum:
**Description**
As I created an 3D example in #237 ,  errors occurred during balancing. The example contains face, edge (2x) and corner (1x) connections.

Two types of errors occur using different maximum refinement level.
Case 1 - line 1042:
```
[libsc 0] Abort: Unreachable code
[libsc 0] Abort: ../p4est-1/src/p4est_algorithms.c:1042
[libsc 0] Abort: Obtained 10 stack frames
```
Case 2 - line 1062:
```
[libsc 0] Abort: Assertion 'p4est_quadrant_is_ancestor (tq, s) || (p4est_quadrant_is_equal (tq, s) && tq->level == P4EST_QMAXLEVEL)'
[libsc 0] Abort: ../p4est-1/src/p4est_algorithms.c:1062
[libsc 0] Abort: Obtained 10 stack frames
```
**To Reproduce**
The parameters below have been tested.
- no mpi, max_level = 2
Case 1 - line 1042
- no mpi, max_level = 3, 4, 5
Case 2 - line 1062
- with mpi, no_of_process = 1, max_level = 2
Case 1 - line 1042
- with mpi, no_of_process = 1, max_level = 3, 4, 5
Case 2 - line 1062

- with mpi, no_of_process = 2, 3, max_level = 2, 3, 4, 5
Case 2 - line 1062

Errors persist with P8EST_CONNECT_FULL/FACE/EDGE.

cburstedde:
Thanks for the report!

cburstedde:
It will make sense to integrate #101 on the way to resolving this issue.

tim-griesbach:
@hannesbrandt  and I checked the issue and we think that the reason for the reported assertion failure is connected  to `p4est_comm_neighborhood_owned`.  The assertion in `p4est_tree_compute_overlap` fails because the binary search does not find a result due to missing quadrants in the `borders` array determined in `p4est_balance` and passed to `p4est_balance_response`. In `p4est_tree_compute_overlap` it is assumed that the tree first and last descendants can be used to check for overlap with `borders`.

 The `borders` array is determined in `p4est_balance` in particular using `p4est_comm_neighborhood_owned`. Here, the problem is that some quadrants of the tree boundary are skipped, which violates the assumption in  `p4est_tree_compute_overlap`. The decision for skipping is made by calling `p4est_comm_neighborhood_owned`.

[...]

PR #248, opened 2023-10-10
example: two crashed examples

SAutum:
# add two crashed examples in balance

Following up on issue #238  .

Proposed changes:
- add a 2d bowtie example in the _balance_ example
- add a 2d drop example in the _balance_ example
- some minor changes along the way to fix warnings in doc generation


cburstedde:
Thanks; if you could add a line to the doc/release_notes.txt file, and another retroactively for your previous #236?

cburstedde:
Thanks! I'm thinking that the 2D and 3D corner examples have much code in common. Do you think you could eliminate the 3.c file to the usual 2-liner we use in the library files and work with `#ifdef P4_TO_P8` in the 2.c example file?

SAutum:
> Thanks! I'm thinking that the 2D and 3D corner examples have much code in common. Do you think you could eliminate the 3.c file to the usual 2-liner we use in the library files and work with `#ifdef P4_TO_P8` in the 2.c example file?

OK. Will do.

cburstedde:
> > Thanks! I'm thinking that the 2D and 3D corner examples have much code in common. Do you think you could eliminate the 3.c file to the usual 2-liner we use in the library files and work with `#ifdef P4_TO_P8` in the 2.c example file?
> 
> OK. Will do.

Thanks; please note that I have pushed a small update to this branch in the meantime.

example/balance/balance_corner2.c
```diff
@@ -0,0 +1,160 @@
...
+#include <p4est_bits.h>
+#include <p4est_extended.h>
+#include <p4est_vtk.h>
+#else
+#include <p8est_bits.h>
+#include <p8est_extended.h>
+#include <p8est_vtk.h>
+#endif
+
+typedef struct
```
cburstedde, line 35: This example contains some historic items for demonstration purposes.
The user_data may be deleted here, and data_size hardcoded as 0 below.

example/balance/balance_corner2.c
```diff
@@ -0,0 +1,160 @@
...
+  int                 mpirank;
+}
+mpi_context_t;
+
+/* refinement level initialization */
+static int          refine_level = 0;
+
+/* connectivity initialization function */
+static void
+init_fn (p4est_t * p4est, p4est_topidx_t which_tree,
```
cburstedde, line 54: This init function is for demonstration purposes and fine in itself.
This example is a good candidate for maximum simplification, so deleting init and hardcoding it below as NULL would be fine.

example/balance/balance_corner2.c
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  /* assign variables based on configuration */
+  refine_level = atoi (argv[1]);
+
+  /* create connectivity and forest structures */
+  geom = NULL;
+#ifndef P4_TO_P8
+  connectivity = p4est_connectivity_new_bowtie ();
+  p4est = p4est_new_ext (mpi->mpicomm, connectivity, 15, 0, 0,
```
cburstedde, line 127: The new_ext function is translated and can live outside the #ifdefs. We had historically chosen the parameter 15 for min_quadrants, but that makes the refinement partition-dependent, which we later found undesirable. This value can be set to 0. Conversely, the fill_uniform parameter would need to be set to 1 to make sure that the refinement looks the same independent of the number of cores used.

example/balance/balance_corner2.c
```diff
@@ -0,0 +1,160 @@
...
+                         sizeof (user_data_t), init_fn, geom);
+#else
+  connectivity = p8est_connectivity_new_drop ();
+  p4est = p8est_new_ext (mpi->mpicomm, connectivity, 15, 0, 0,
+                         sizeof (user_data_t), init_fn, geom);
+#endif
+  p4est_vtk_write_file (p4est, geom, "test");
```
cburstedde, line 134: We may consistently produce output with file names, after calling the respective algorithms, like
P4EST_STRING "_corner_new",
P4EST_STRING "_corner_refine",
P4EST_STRING "_corner_balance",
which will allow us to write the VTK outside the dimension #ifdefs.

SAutum: Deleted. The statement serves a test on the p4-to-p8 functionality.

cburstedde: I don't mind writing a VTK after new, after refine, after balance, and after partition.
With the name convention as above, this should all work without conflict:
with P4EST_STRING, the vtk_write_file function can be called outside the #ifdef.

cburstedde:
Thanks, I have pushed on top as discussed in our video meeting.

PR #249, opened 2023-10-12
sphere2d geometry

ljcarlin:
# Draft PR for sphere2d geometry

Proposed changes: Add sphere geometry for use with the 6-tree cubed connectivity. Add configuration to use this geometry in example/simple/simple2.c


cburstedde:
Thanks! If you were to base the PR on the develop branch that would be fitting for code that is feature-oriented as opposed to fundamental/formal.

I'd like to think that in 2D we generally ignore the input z coordinate. We may double-check that this is the case on the latest develop branch (I just pushed a fix to that end) and document it accordingly under \file p4est_geometry.h.

ljcarlin:
Whoops. I will redo a version based on develop.

The unnecessary z coordinate was one of the things I wanted to ask about today!

ljcarlin:
Closed in favour of https://github.com/cburstedde/p4est/pull/250

PR #250, opened 2023-10-12
sphere2d geometry

ljcarlin:
# PR for sphere2d geometry

Fixed this to be based on develop instead of master.

Proposed changes: Add sphere geometry for use with the 6-tree cubed connectivity. Add configuration to use this geometry in example/simple/simple2.c. Add documentation for p4est_geometry

### DIFF
--- a/example/simple/simple2.c
+++ b/example/simple/simple2.c
@@ -42,7 +42,7 @@
  *        o rotwrap   Refinement on the unit square with weird periodic b.c.
  *        o circle    Refinement on a 6-tree donut-like circle.
  *        o drop      Refinement on a 5-trees geometry with an inner hole.
- *        o icosahedron   Refinement on the sphere
+ *        o icosahedron   Refinement on the icosahedron sphere with geometry.
  *        o shell2d       Refinement on a 2d shell with geometry.
  *        o disk2d        Refinement on a 2d disk with geometry.
  *        o bowtie    Refinement on a 2-tree bowtie domain.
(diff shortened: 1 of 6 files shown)


cburstedde:
Thanks this looks great!
Were you planning to add documentation/extra code to make clear that abc[2] is never accessed?

ljcarlin:
I haven't finished cleaning up the documentation yet, so I will add that bit too.

 I was having trouble making the function we discussed public, so I've been working on the GMT stuff in the meantime.

cburstedde:
> I haven't finished cleaning up the documentation yet, so I will add that bit too.
> 
>  I was having trouble making the function we discussed public, so I've been working on the GMT stuff in the meantime.

We can, for this PR, leave the function scoped as is and worry about it separately.

ljcarlin:
Sounds good!

I finished including all the documentation I intended, so it should be good to go. 

There were a few different terms being used to describe local coordinates. I thought the clearest was "tree-local reference coordinates" so I've used it to replace other terms like "AMR space".

I also added assertions so that `p4est_geometry_connectivity_X` doesn't segfault if given a geometry without vertex information.

ljcarlin:
Sorry for the weird force push. I forgot to rebase upstream commits before pushing. Hope that didn't break anything...

cburstedde:
No problem! Looking forward to iron out the last remaining doxygen warnings, and possibly getting the p8est_geometry docs aligned.

ljcarlin:
Resolved all the doxygen geometry errors except this one coming from another file:

`p4est/src/p4est_vtk.h:92: warning: unable to resolve reference to 'p4est_geometry_t' for \ref command`

It's coming from an edgecase that doxygen is not handling well.

ljcarlin:
Okay, all doxygen warnings are fixed!

cburstedde:
Thanks so much, I've tweaked a bit further, some very old content, too.
I've gone through a couple conventions (running p4estindent on the files, removing superfluous spaces in places).

PR #256, opened 2023-10-27
Update libsc (with updated CMake logic)

cburstedde:
# Update libsc with updated CMake logic

I see a CMake failure on the CI. Han this be fixed?


cburstedde:
@scivision: wondering if there are any implications on the p4est CMake by latest libsc. Any ideas? Thanks!

cburstedde:
I force-pushed to make libsc test ok with valgrind. Still wondering what this is about. :)

cburstedde:
This branch refers to the latest libsc with @scivison's PR merged. Ideally, development on top of this branch might get the CI to work with the current module setting.

cburstedde:
I've bumped the test_develop branch with the available code from that other repository.
The test are now going through again! I'm just suspecting that the SC_HAVE_ZLIB and P4EST_HAVE_ZLIB defines end up having different values for some of the CI configurations.

cburstedde:
One more thing that CMake may not yet be figuring out in all configurations is the convention for qsort_r: GNU or BSD. In particular, the defines `SC_HAVE_QSORT_R`, `SC_HAVE_BSD_QSORT_R` would need to be set or unset accordingly.

PR #258, opened 2023-11-22
Feature gmt sphere

ljcarlin:
# gmt sphere Draft PR 

This builds another example on top of the search based refinement in feature-gmt. More specifically it refines the spherical geometry based on geodesics on the surface of the sphere. Geodesics are read in from a csv where each line `phi1,theta1,phi2,theta2` represents the two endpoints of a geodesic in spherical coordinates.

The options handling is not very elegant at the moment so I will fix this. I will also clean up the documentation to better explain the workflow of preparing a dataset for use.


ljcarlin:
I cleaned up the options handling and added documentation for how to use the sphere model. 

I wasn't sure if I should also include the python script I used to extract line segment data from .shp files.

PR #259, opened 2023-11-22
Geometry API change

ljcarlin:
The function `p4est_geometry_connectivity_X` is a useful building block for defining custom geometries. I have made it public so that users may take advantage of this.

I also fixed a typo in the geometry file and updated some misleading documentation.


cburstedde:
Thanks!

Issue #244, opened 2023-10-02
Edge-hanging corner neighbors in the p8est_mesh_t


hannesbrandt:
As pointed out by @scottaiton  (https://github.com/cburstedde/p4est/pull/235) the `p8est_mesh_t` does not provide information about same size corner-neighbors across a hanging edge, since the corresponding entries of `quad_to_corner` are set to -1.
We could provide the hanging corner neighbors using `corner_offset`, `corner_quad` and `corner_corner`. The `quad_to_corner` array would then index into local_num_quadrants + local_num_ghosts + [0 .. local_num_corners - 1]. We could use the entries of `corner_corner` to encode information similar to `edge_edge`, e.g. by setting `corner_corner` to 0..7 for non-hanging neighbors and to 8..15 for hanging corner neighbors.
Thereby, we have access to the corner neighbors across hanging edges and maintain the possibility to distinguish between hanging and non-hanging neighbors.

Furthermore, there is an open todo in the documentation of the `p8est_mesh_t`
```
TODO: In case of an inter-tree neighbor relation in a brick-like
 *       situation (one same-size neighbor, diagonally opposite edge/corner),
 *       use the same encoding as for edges/corners within a tree.
```
This might be a good opportunity to resolve it as well.

cburstedde:
> As pointed out by @scottaiton (#235) the `p8est_mesh_t` does not provide information about same size corner-neighbors across a hanging edge, since the corresponding entries of `quad_to_corner` are set to -1. We could provide the hanging corner neighbors using `corner_offset`, `corner_quad` and `corner_corner`. The `quad_to_corner` array would then index into local_num_quadrants + local_num_ghosts + [0 .. local_num_corners - 1]. We could use theentries of `corner_corner` to encode information similar to `edge_edge`, e.g. by setting `corner_corner` to 0..7 for non-hanging neighbors and to 8..15 for hanging corner neighbors. Thereby, we have access to the corner neighbors across hanging edges and maintain the possibility to distinguish between hanging and non-hanging neighbors.

I'd agree to add this case with the necessary information. I'm wondering whether we'd need to use the 8..15 values, since for edge_edge we use it for orientation, where here we'd use it for flagging hanging/non-hanging context. Since this information can, if necessary, be looked up from child id and corner number, it may suffice to state the formula/expression in the documentation and use 0..7 values only.

[...]

PR #262, opened 2023-12-07
Edge-hanging corners in mesh

hannesbrandt:
# p4est_mesh: add corner neighbors across coarse edges

Following up on issue https://github.com/cburstedde/p4est/issues/244 and the test proposed by @scottaiton in https://github.com/cburstedde/p4est/pull/235.
We introduce a `p4est_mesh_params_t` that contains all mesh-creation parameters with the possibility to add further parameters in the future. The `p4est_mesh_t` now contains a `p4est_mesh_params_t` instance to store the parameters it was created with.
In 3D, we add a new `edgehanging_corners` flag to the set of parameters. When set to true, quadrants that are corner neighbors across a coarse edge are stored in the same format as all other corner-neighbor relations of the mesh. If set to false, the mesh is created as usual.
Furthermore, we add `p4est_mesh_new_params` which creates a mesh according to a passed `p4est_mesh_params_t` instance. A default parameter instance can be created with `p4est_mesh_params_new` and destroyed with `p4est_mesh_params_destroy`.
Apart from this, we remove a few unreachable lines of code in `mesh_iter_edge` and add some documentation in `p4est_mesh.h`. Especially, `p4est_mesh_get_neighbors` is now marked as outdated.


src/p4est_mesh.c
```diff
@@ -162,6 +171,16 @@ mesh_edge_process_inter_tree_edges (p8est_iter_edge_info_t * info,
   equads = P4EST_ALLOC (p4est_locidx_t, nAdjacentQuads);
   eedges = P4EST_ALLOC (int8_t, nAdjacentQuads);
 
+  ccorners = NULL;
+  cquads = NULL;
+  if (mesh->params.edgehanging_corners == 1 &&
```
cburstedde, line 176: Please check booleans with if (a) or if (!a). The C standard considers any nonzero as true.

src/p4est_mesh.c
```diff
@@ -554,56 +602,22 @@ mesh_iter_edge (p8est_iter_edge_info_t * info, void *user_data)
     return;
   }
   if (cz == 2) {
+    P4EST_ASSERT (info->tree_boundary);
```
cburstedde, line 605: Good catch!

src/p4est_mesh.c
```diff
@@ -1136,6 +1212,36 @@ p4est_mesh_memory_used (p4est_mesh_t * mesh)
   return all_memory;
 }
 
+static void
+mesh_params_init (p4est_mesh_params_t * params)
```
cburstedde, line 1216: I'd be suggesting to work with static parameters. Then we wouldn't need the new and destroy functions, just p4est_mesh_params_init, which first memsets the whole structure to zero and then writes defaults into it.

src/p4est_mesh.c
```diff
@@ -1147,6 +1253,23 @@ p4est_mesh_t       *
 p4est_mesh_new_ext (p4est_t * p4est, p4est_ghost_t * ghost,
                     int compute_tree_index, int compute_level_lists,
                     p4est_connect_type_t btype)
+{
+  p4est_mesh_params_t parameters;
+
```
cburstedde, line 1258: If we call p4est_mesh_params_init (&mesh->parameters) we avoid duplicate settings.

cburstedde: You're right; I had a slightly different angle. The `mesh->params = *params;` is fine.

I was thinking of using after `/* initialize parameter struct to pass to mesh_new_params */`
something like `p4est_mesh_init_params (&parameters)` and thereafter only overwriting the values that are not default. Would that be safer or do you prefer the current approach?

As a side note, shall we rename any leftover `parameters` to `params` for consistency?

hannesbrandt: That's a good point, initializing first with `p4est_mesh_init_params` ensures that new additions to the `mesh_params_t` do not lead to unexpected behaviour in `mesh_new_ext`. I've changed this and renamed `parameters` to `params` as well.

cburstedde: Looks very nice!

PR #263, opened 2023-12-08
Feature gmt sphere preprocessing

ljcarlin:
This is the preprocessing component for my sphere geodesic refinement model. Usage documentation is included at the top of the file.


example/gmt/Makefile.am
```diff
@@ -12,6 +12,14 @@ example_gmt_p4est_gmt_SOURCES = \
 
 LINT_CSOURCES += \
         $(example_gmt_p4est_gmt_SOURCES)
+
+bin_PROGRAMS += example/gmt/p4est_sphere_preprocessing
+
+example_gmt_p4est_sphere_preprocessing_SOURCES = example/gmt/sphere_preprocessing.c
+
+LINT_CSOURCES += \
```
cburstedde, line 20: We're in the process of removing everything related to LINT. It would be ok to remove it for this example already.

example/gmt/gmt_models.h
```diff
@@ -73,6 +73,18 @@ p4est_gmt_model_latlong_params_t;
 p4est_gmt_model_t  *p4est_gmt_model_latlong_new
   (p4est_gmt_model_latlong_params_t * params);
 
+/** Represents a segment of a geodesic in the sphere model.
```
cburstedde, line 76: Would there be a simple way to have the doxygen parser include this .h file?

example/gmt/sphere_preprocessing.c
```diff
@@ -0,0 +1,525 @@
...
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with p4est; if not, write to the Free Software Foundation, Inc.,
+  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+*/
+
+/** \file sphere_preprocessing.c
```
cburstedde, line 25: Thanks; nice description. If we were to include this program as one of our examples using the doxygen system, this can be done similarly like @SAutum has done it for `example/simple`.

To have the user run the program, would you like to include a small .csv input file that makes a friendly hello world type picture?

As for the main routine, would it make sense to optionally provide a second command line argument as name of the output file when it is desired to be written?

example/gmt/sphere_preprocessing.c
```diff
@@ -0,0 +1,525 @@
...
+
+  /* Convert to radians */
+  phi = angular[0] * M_PI / 180.0;
+  theta = angular[1] * M_PI / 180.0;
+
+  xyz[0] = sin (theta) * cos (phi);
+  xyz[1] = sin (theta) * sin (phi);
+  xyz[2] = cos (theta);
+
+  inf_norm = fmax (fabs (xyz[0]), fmax (fabs (xyz[1]), fabs (xyz[2]))) * 2.0;
```
cburstedde, line 64: `fmax` is not necessarily included in a C library. Would `SC_MAX` be an option, optionally using a temporary variable to avoid duplicate computation? Also, a performance extremist would store the inverse once and use `*=` three times.

example/gmt/sphere_preprocessing.c
```diff
@@ -0,0 +1,525 @@
...
+  xyz[0] /= inf_norm;
+  xyz[1] /= inf_norm;
+  xyz[2] /= inf_norm;
+  return;
+}
+
+/**
+ * Which face does a given point belong to
```
cburstedde, line 72: This is an illustrative example matching the specific geometry. Would it make sense to cross-reference using doxygen?

example/gmt/sphere_preprocessing.c
```diff
@@ -0,0 +1,525 @@
...
+    return 0;
+  if (fabs (xyz[2] - 0.5) < SC_EPS)
+    return 3;
+  return -1;                    /* This should not happen */
+}
+
+/**
+ * coordinate transformation from the surface of cube [-0.5,0.5]^3
+ * to tree-local coordinates. This is the inverse of 
+ * p4est_geometry_cubed_X
```
cburstedde, line 99: Might we use \ref to cross-link wherever we can?

example/gmt/sphere_preprocessing.c
```diff
@@ -0,0 +1,525 @@
...
+ *
+ * \param[in]  xyz  cartesian coordinates in physical space
+ * \param[in]  which_tree face of cube
+ * \param[out] rst  coordinates in AMR space : [0,1]^3
+ *
+ */
+static void
```
cburstedde, line 106: Would it make sense to P4EST_ASSERT the ranges of input arguments?

example/gmt/sphere_preprocessing.c
```diff
@@ -0,0 +1,525 @@
...
+    break;
+  case 4:
+    rst[0] = xyz[0] + 0.5;
+    rst[1] = xyz[2] + 0.5;
+    break;
+  case 5:
+    rst[0] = xyz[1] + 0.5;
+    rst[1] = xyz[2] + 0.5;
+    break;
+  default:
```
cburstedde, line 138: We may SC_ABORT_NOT_REACHED whenever we encounter illegal switch values.

example/gmt/sphere_preprocessing.c
```diff
@@ -0,0 +1,525 @@
...
+  /* Compute the determinant by Laplace expansion along the first column */
+  det_A = 0;
+  for (int r = 0; r < 3; r++) {
+    det_A += A[r][0] * cofactor[r][0];
+  }
+
+  /** If the determinant is zero then the line segment is parallel to the 
+   * cone. We count this case as not intersecting.
+  */
+  if (det_A < SC_EPS) {
```
cburstedde, line 189: Sometimes SC_1000_EPS is more robust. We need a well defined plan B in that case.

example/gmt/sphere_preprocessing.c
```diff
@@ -0,0 +1,525 @@
...
+    {{-0.5, -0.5, -0.5}, {-0.5, -0.5, 0.5}},    /* 2,4 edge */
+    {{-0.5, -0.5, 0.5}, {0.5, -0.5, 0.5}},      /* 3,4 edge */
+    {{0.5, -0.5, 0.5}, {0.5, 0.5, 0.5}},        /* 3,5 edge */
+    {{0.5, -0.5, -0.5}, {0.5, -0.5, 0.5}}       /* 4,5 edge */
+  };
+
+  detected = cone_line_intersection (xyz1, xyz2, edge_endpoints[edge][0],
+                                     edge_endpoints[edge][1], p_intersect);
+
+  if (detected) {
```
cburstedde, line 263: In general, a couple more standard comments on the math will be helpful to the reader throughout. I know that so far p4est is not a stellar example on extensive commenting, but please feel free to improve wherever you can.

example/gmt/sphere_preprocessing.c
```diff
@@ -0,0 +1,525 @@
...
+        geodesics[n_geodesics].p2x = rst2[0];
+        geodesics[n_geodesics].p2y = rst2[1];
+
+        n_geodesics++;
+      }
+    }
+  }
+
+  /* Free extra capacity */
+  geodesics =
```
cburstedde, line 438: Nice being tight on memory requirements.

example/gmt/sphere_preprocessing.c
```diff
@@ -0,0 +1,525 @@
...
+    if (input == NULL) {
+      P4EST_GLOBAL_LERROR ("Could not find input file\n");
+      P4EST_GLOBAL_LERROR (usage);
+      progerr = 1;
+    }
+  }
+
+  /* split geodesics into segments */
+  if (!progerr) {
+    progerr = compute_geodesic_splits (&n_geodesics, &geodesics, input);
```
cburstedde, line 483: Adding an error message would help the user out just a bit more.

example/gmt/sphere_preprocessing.c
```diff
@@ -0,0 +1,525 @@
...
+  }
+
+  /* close input file */
+  if (input != NULL) {
+    fclose (input);
```
cburstedde, line 488: Checking the return of every fclose will be nice, even though we'd only print a message since progerr is already true.

cburstedde: It's a good thing not checking for progerr here and closing depending exactly when required.

example/gmt/sphere_preprocessing.c
```diff
@@ -0,0 +1,525 @@
...
+  }
+
+  /* open output file for writing */
+  if (!progerr) {
+    output = fopen ("geodesics", "w");
+    if (output == NULL) {
+      progerr = 1;
+      P4EST_GLOBAL_LERROR ("File open fail\n");
```
cburstedde, line 496: Wrt. my above comment on specifying the file name on the command line, the name may be repeated in the message to be more explicit to the user. There is also the option to use ferror and/or errno to be even more closely behaving like any unix command line tool.

cburstedde: All these extra checks may warrant to introduce more helper functions to avoid code redundancy.

example/gmt/sphere_preprocessing.c
```diff
@@ -0,0 +1,525 @@
...
+    }
+  }
+
+  /* write number of geodesic segments */
+  if (!progerr) {
+    nwritten = fwrite (&n_geodesics, sizeof (int), 1, output);
+    if (nwritten != 1) {
```
cburstedde, line 503: Making this an exercise how to best check for fwrite errors according to its documentation would be instructive.

example/gmt/sphere_preprocessing.c
```diff
@@ -0,0 +1,525 @@
...
+                       n_geodesics, output);
+    if (nwritten != n_geodesics) {
+      progerr = 1;
+      P4EST_GLOBAL_LERROR ("File write fail: geodesics\n");
+    }
+  }
+
+  /* finished writing geodesics */
+  if (output != NULL) {
+    fclose (output);
```
cburstedde, line 521: See above. Thanks so far, this is looking very good.

example/gmt/gmt_models.h
ljcarlin: We could add it to the file list. I think it makes sense to do this for the rest of the examples too. Currently the file list only tracks /src/, but I don't see why it shouldn't include all of /example/ as well.

example/gmt/sphere_preprocessing.c
ljcarlin: >Thanks; nice description. If we were to include this program as one of our examples using the doxygen system, this can be done similarly like @SAutum has done it for example/simple.

Good idea. In my second PR there's more complete usage documentation that I think would belong in that location. I will include it when we get to that PR.

>To have the user run the program, would you like to include a small .csv input file that makes a friendly hello world type picture?

Can do!

>As for the main routine, would it make sense to optionally provide a second command line argument as name of the output file when it is desired to be written?

It complicates the options handling a bit, but I think there is probably a use case for it so I will do it.

example/gmt/sphere_preprocessing.c
ljcarlin: Do you mean that I should reference the geometry here, or that in the documentation for the geometry this should be mentioned as an example?

example/gmt/sphere_preprocessing.c
ljcarlin: Good suggestion! I just realised I actually fail these assertions due to numerical round-off issues.

I'll have a think about the best way to address this.

example/gmt/gmt_models.h
cburstedde: It is intentional not to have the example .c files in the list. They should be commented profusely with `/* */`.
For example .h files it's a different matters, and here I would agree to do this in a hand-picked way.

example/gmt/sphere_preprocessing.c
cburstedde: I hope this would be benign enough, when not given we do not write any output.

On a related note, would it make sense to have this example do nothing on any MPI rank > 0? In future incarnations of course, parallel operation will be implemented.

example/gmt/sphere_preprocessing.c
cburstedde: I'm referring to using /** \ref <existing keyword> */ in comments where sensible.

cburstedde:
Thanks so far! Sorry for the micro-comments. You may use your own better judgement.

example/gmt/sphere_preprocessing.c
ljcarlin: I just copied the way sc_fwrite error checks. Do you mean using something like `explain_fwrite`?

example/gmt/sphere_preprocessing.c
ljcarlin: I switched to SC_1000_EPS. It didn't make any difference in my real dataset anyway.

ljcarlin:
> Thanks so far! Sorry for the micro-comments. You may use your own better judgement.

They're all very helpful, thanks!

example/gmt/gmt_models.h
ljcarlin: The file list is set up to only parse .h files, so in theory we could just let it parse all the .h files in /example/. Or just hand pick ones to include.

There's also a separate examples list which doxygen gives its own page at /doc/html/examples.html. Currently the list is very outdated. I could add it to that list though and work on improving the list in the future.

example/gmt/sphere_preprocessing.c
ljcarlin: > On a related note, would it make sense to have this example do nothing on any MPI rank > 0? In future incarnations of course, parallel operation will be implemented.

Sure.

example/gmt/gmt_models.h
cburstedde: We might sit down with @SAutum to discuss how to go forward with documenting examples. This will be quite relevant down the road.

example/gmt/sphere_preprocessing.c
cburstedde: We don't need to overdo it. The program should just behave in a way that you personally like on whatever file error condition that occurs.

PR #269, opened 2023-12-22
CMake: check that Zlib works

scivision:
* verify SC has a good zlib (in case old libsc that didn't have full check)
* allow top-level project to override default install dir

this does not depend on the pending libsc PR

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,7 +139,7 @@ jobs:
         cd sc
         git fetch --tags
         # git checkout v2.8.5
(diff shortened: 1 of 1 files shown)


cburstedde:
Thanks! There seems to be one remaining CI issue remaining, which I'd like to wait out.

cburstedde:
Thanks! I've updated everything available in libsc and registered the latest submodule with p4est/develop. You should be in good shape to rebase this PR history sans sc updates. Looking forward to pull it in then!

PR #271, opened 2024-01-07
cmake: P4EST_BUILD_TESTING to allow parent project control

scivision:
by using a more specific name like "P4EST_BUILD_TESTING" instead of "BUILD_TESTING" this allows parent projects like Forestclaw to select which tests to build and run. This avoids outdated test dependencies causing a top-level project to fail build/test.

Like libsc, since CDash isn't used, there isn't a need for the overhead of `include(CTest)`.

This does _not_ depend on libsc PR just made.


cburstedde:
Cool thanks!

PR #272, opened 2024-01-15
Feature soversion

cburstedde:
# Read SOVERSION from configuration file.

We place the library API version by the libtool convention in one global configuration file.
This is sourced from both the autoconf and cmake build processes.

In passing increase the CI runtime for CMake on MacOS from 20 to 25 minutes.

### DIFF
--- a/.github/workflows/ci_cmake.yml
+++ b/.github/workflows/ci_cmake.yml
@@ -100,7 +100,7 @@ jobs:
   mac:
     runs-on: macos-latest
     name: CMake build on MacOS
-    timeout-minutes: 20
+    timeout-minutes: 25
 
     strategy:
       matrix:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,7 +30,7 @@ include(cmake/compilers.cmake)
 
 add_library(p4est)
 set_property(TARGET p4est PROPERTY EXPORT_NAME P4EST)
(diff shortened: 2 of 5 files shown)


cburstedde:
We may need to add at least this one new file in config/ to cpack, and the rest of the directory too if not already included.

tim-griesbach:
> We may need to add at least this one new file in config/ to cpack, and the rest of the directory too if not already included.

In `cmake/install.cmake`, we have
```
# not .gitignore as its regex syntax is more advanced than CMake
set(CPACK_SOURCE_IGNORE_FILES .git/ .github/ .vscode/ _CPack_Packages/
${CMAKE_BINARY_DIR}/ ${PROJECT_BINARY_DIR}/
Makefile.in
aclocal.m4
autom4te.cache/
build/
build-aux/ar-lib
build-aux/compile
build-aux/config.guess
build-aux/config.sub
build-aux/depcomp
build-aux/install-sh
build-aux/ltmain.sh
build-aux/missing
build-aux/test-driver
config/libtool.m4
config/ltoptions.m4
config/ltsugar.m4
config/ltversion.m4
config/lt~obsolete.m4
configure
src/sc_config_autotools.h.in
scindent
scspell
)
``` 
Therefore, I think that the new file in `config/` should be already included.

tim-griesbach (commented):
This looks good to me!

PR #273, opened 2024-01-17
cmake: simplify by not using FetchContent

scivision:
using add_subdirectory() eliminates unnecessary overhead of FetchContent. This allows users to switch mpi on/off more easier with the --fresh CMake option, rather than having to manually delete the build directory.

This doesn't change anything from end-users perspective other than that it's easier to use with less files in the build directory.


cburstedde:
Thanks! Looking forward to further integration.

Is there a way to eliminate the transient CI failures? They have been occurring with almost every update over the last couple weeks. I'm just puzzled since that never happens with autoconf. It would be great to be able to rely on this for the CMake CI as well.

cburstedde:
Merging now!

PR #274, opened 2024-01-18
CI: improve stability by increasing too-low time limits

scivision:
the runtime of tests on GitHub Actions and CI systems in general
has considerable variance due to different virtual worker hardware
and system loads.

The individual CTest test timeouts increase from 10 => 60 seconds.
The overall job time limit increases to 60 minutes. The only reason
for the overall time limit is so that a hung download or MPI job
doesn't sit "running" for a couple days.

GitHub Actions now has 4-core CI runners, so use all of them by default.

also, include GCC-13 in tests.

also, use the latest Actions for checkout and upload-artifact

Example of nuisance transient timeout failure: https://github.com/cburstedde/p4est/actions/runs/7562394485/job/20592780748)

### DIFF
--- a/.github/workflows/ci_cmake.yml
+++ b/.github/workflows/ci_cmake.yml
@@ -16,7 +16,8 @@ on:
     types: [published]
 
 env:
-  CTEST_PARALLEL_LEVEL: 3
+  CTEST_PARALLEL_LEVEL: 4
+  CMAKE_BUILD_PARALLEL_LEVEL: 4
   CTEST_NO_TESTS_ACTION: "error"
 
 jobs:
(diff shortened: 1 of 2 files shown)


cburstedde:
> @cburstedde GCC-13 was added in this PR, it errors on the following "maybe-uninitialized" code. Is this a false positive i.e. should I use gcc flag "-Wno-maybe-uninitialized"?

Thanks for the report! Please keep the warnings alive. This one is fixed on latest develop branch.

cburstedde:
I see, the CI is failing due to warnings turned into errors. Should go through based on latest develop.

PR #275, opened 2024-01-22
file_info: Fixes and simplifications

tim-griesbach:
# file_info: Fixes and simplifications

Proposed changes: Analogously to the ForestClaw PR https://github.com/ForestClaw/forestclaw/pull/301, we introduce the function `p4est_file_info_cleanup` to simplify the error management in the function `p4est_file_info`. In addition, there are some fixes of the error management in `p4est_file_info`.


cburstedde:
Thanks for checking in!

PR #276, opened 2024-01-23
Doxygen documentation page - Installation

SAutum:
# Documentation page - Installation
This PR is intended to assist users in setting up their builds with ease. 

Proposed changes:
- Create an installation guide page.
- Integrate the information collected from various sources, including the homepage, codes, and GitHub repository.


SAutum:
Thanks for the suggestion, I'll update the documents accordingly

cburstedde:
Sure; I was thinking:
 - add what you like under the Debian section
 - describe most of the build assuming to begin from a tarball without configure
 - add the first steps from the build from source and then suggest to proceed as with the tarball build

cburstedde:
Thanks!

PR #279, opened 2024-01-24
CMake force cache

scivision:
Note: for this to have the desired effect depends on https://github.com/cburstedde/p4est/pull/280
However, #280 had unrelated C source code changes to libsc that broke p4est.

Correct handling of CMake cache variables that may change with user options.

Now fixed in this PR, but previously, switching any of these options resulted in inconsistent, errant project state:

* enable_p6est
* enable_p8est
* mpi


cburstedde:
Thanks, we should have made sufficient sc updates that this will apply fine!

PR #283, opened 2024-01-26
Partition for coarsening in wrap parameters

hannesbrandt:
We add a `partition_for_coarsening` option to the wrap parameters and store a `p4est_wrap_params_t` in the wrap.

- We move several existing parameters, like `hollow` or `coarsen_delay` from the `p4est_wrap_t` into a `p4est_wrap_params_t` instance, which now is a member of the wrap.
- Additionally, we add the `partition_for_coarsening` parameter, which is passed to the calls of `p4est_partition_ext` in `p4est_wrap_partition`.
- We remove `initial_level` from the `wrap_t`, as it is only needed for creating a p4est based on a connectivity and a wrap may also be created for an already existing p4est without providing an initial level.
- For convenience, we introduce `p4est_wrap_new_p4est_params` to create a wrap of an existing p4est based on an input parameter instance. This function is called both by `p4est_wrap_new_p4est` and `p4est_wrap_new_params`.

Currently, the `btype` of the wrap is taken from the `p4est_mesh_params_t` instance that is part of the `p4est_wrap_params_t`. While this ensures consistent connect types across the different structs, the repeated calls to `wrap->params->mesh_params->btype` seem slightly counterintuitive, so it might make sense to add a duplicate `btype` to the `wrap_params_t`.


hannesbrandt:
With the latest commits the internal mesh creation in `wrap_{adapt,partition,hollow}` uses the mesh parameters which are stored in the wrap parameters. So, the mesh parameters specified by the user are now used throughout the whole wrap workflow and not only for `new_params`.
Additionally, I have added `wrap_set_partitioning` for consistency and convenience.

cburstedde:
Thanks for this significant update! We are breaking the wrap API on purpose for the next version of the code, so we might very clearly state this in the release notes. We can do that when we also remove mesh_get_neighbors (let's just not forget). I'm merging now to get some coverage for the new code.

PR #285, opened 2024-02-06
Move bsearch function up

tim-griesbach:
# Move bsearch function up

Proposed changes: The function `p4est_bsearch_partition` was placed between `p4est_transfer_custom` and `p4est_transfer_begin`. The function is moved to the top of `p4est_communication.{c,h}` and of `p8est_communication.h` to make the ordering of the functions more consistent.


cburstedde:
Thanks!

PR #287, opened 2024-02-19
CI for checking if PR changes submodule

mkirilin:
This CI runs on pull request and checks whether the PR attempts to change reference to sc submodule. If so, it exiting with error, turns a PR into a draft PR, and leaves a comment:

"Submodule has been changed by this PR. This is forbidden by the repository policy. Your PR is turned into a draft PR."

Otherwise, it exits with success and does nothing.

### DIFF
--- a/.github/workflows/ci_submodule.yml
+++ b/.github/workflows/ci_submodule.yml
@@ -1,5 +1,4 @@
-name: CI for libsc submodule checking
-
+name: CI for libsc check
 on:
   pull_request:
 
--- a/.github/workflows/ci_submodule.yml
+++ b/.github/workflows/ci_submodule.yml
@@ -12,9 +12,11 @@ jobs:
       uses: actions/checkout@v3
       with:
         submodules: true
-        token: ${{ github.token }}
 
     - name: Check if submodule is changed
+      env:
+        TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        PR_NODE_ID: ${{ github.event.pull_request.node_id }}
       shell: bash
       run: |
         EVENT_PATH="${GITHUB_EVENT_PATH}"
@@ -37,11 +39,11 @@ jobs:
         echo "Check if submodule has been changed on ${PR_BRANCH}"
         CHANGED=`git diff --name-only ${FROM_HASH}...${TO_HASH}`
         if grep "^${INPUT_PATH}$" <<< "${CHANGED}"; then
-          echo "Submodule has been changed by this PR. \
-          It is forbidden by the repository policy. \
+          echo "Submodule has been changed by this PR. \n
+          It is forbidden by the repository policy. \n
           Changing it to be a draft PR..."
-          echo ${{ inputs.token }} | gh auth login --with-token
-          gh api graphql -F id=${{ inputs.pull-request-node-id }}  -f query='
+          echo "${TOKEN}" | gh auth login --with-token
+          gh api graphql -F id="${PR_NODE_ID}"  -f query='
             mutation($id: ID!) {
               convertPullRequestToDraft(input: { pullRequestId: $id }) {
                 pullRequest {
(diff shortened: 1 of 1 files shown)


cburstedde:
Thanks; there may be a simpler way without checking out the submodules.

.github/workflows/ci_submodule.yml
```diff
@@ -0,0 +1,72 @@
...
+    runs-on: ubuntu-latest
+    env:
+      TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      PR_NODE_ID: ${{ github.event.pull_request.node_id }}
+      INPUT_PATH: sc
+    steps:
+    - name: Checkout source code for base
+      uses: actions/checkout@v4
+      with:
+        submodules: true
```
cburstedde, line 18: Would this work with submodules false and just checking the sc hash relative to each p4est subdirectory?

.github/workflows/ci_submodule.yml
```diff
@@ -0,0 +1,72 @@
+name: CI for libsc check
+on:
+  pull_request:
+
+jobs:
+
+  linux-sub:
+    if: github.event.pull_request.draft == false
```
cburstedde, line 8: Collaborators should be allowed to post a non-draft PR.

.github/workflows/ci_submodule.yml
mkirilin: Thanks for the idea. It is fixed in the recent update.

.github/workflows/ci_submodule.yml
mkirilin: Dode.

mkirilin:
UPD: I've simplified the code by checking out the base repository's branch initially without. Fetching is no need anymore.

.github/workflows/ci_submodule.yml
```diff
@@ -0,0 +1,69 @@
...
+        repository: ${{ github.event.pull_request.head.repo.full_name }}
+        ref: ${{ github.event.pull_request.head.ref }}
+        submodules: false
+        path: head
+        
+    - name: Determine hashes to compare
+      shell: bash
+      run: |
+        EVENT_PATH="${GITHUB_EVENT_PATH}"
+        cd "${GITHUB_WORKSPACE}" \
```
cburstedde, line 37: One more comment: do we need this line? It seems we are already in this directory.

.github/workflows/ci_submodule.yml
```diff
@@ -0,0 +1,69 @@
...
+        || error "__Line:${LINENO}__Error: Cannot change directory to Github Workspace"
+        PR=`jq -r ".number" "${EVENT_PATH}"`
+        cd head
+        HEAD_SC_HASH=`git log -n1 sc | awk 'NR==1{ print $2 }'`
+        cd ../base
+        BASE_SC_HASH=`git log -n1 sc | awk 'NR==1{ print $2 }'`
+        echo "HEAD_SC_HASH: $HEAD_SC_HASH, BASE_SC_HASH: $BASE_SC_HASH"
```
cburstedde, line 44: Can we cd back into .. for a well defined situation after this line?

.github/workflows/ci_submodule.yml
mkirilin: You are right, we started in the same solder. It's fixed now.

.github/workflows/ci_submodule.yml
mkirilin: Not after this, because we need to be in a repository to use github API functions later. But I fixed it before we exit the action for consistency.

cburstedde: Ah cool. In that case we may not need the cd (it would have to be && not &, but let's leave it out).

mkirilin: Alright, then I'll undo the changes.

PR #288, opened 2024-02-20
Feature GMT sphere: fixed typo causing segfault

ljcarlin:
Quick fix to a typo in my last PR.

### DIFF
--- a/example/gmt/gmt_models.c
+++ b/example/gmt/gmt_models.c
@@ -494,7 +494,6 @@ p4est_gmt_model_sphere_new (int resolution, const char *input,
   int                 mpireslen;
   char                mpierrstr[sc_MPI_MAX_ERROR_STRING];
   sc_MPI_Offset       mpi_offset;
-  sc_MPI_Status       status;
   const char         *count_mismatch_message =
     "This should only occur when attempting to read "
     "beyond the bounds of the input file. "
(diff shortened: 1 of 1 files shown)


cburstedde:
Thanks! I'm wondering about the error checking of sc_io_read_at_all.

 - The functions should be collective; can you please check whether the error return is synchronized by definition?
 - In the (unlikely?) case that it is not collective, an MPI_Allreduce with a MPI_LOR operator on the retval scales better.

ljcarlin:
> Thanks! I'm wondering about the error checking of sc_io_read_at_all.
> 
> * The functions should be collective; can you please check whether the error return is synchronized by definition?
> * In the (unlikely?) case that it is not collective, an MPI_Allreduce with a MPI_LOR operator on the retval scales better.

My understanding is that all collective functions synchronize errors by definition, so you're correct that the return error does not need to be synchronized. Since count mismatches are not MPI errors they still need to be synchronized. I will replace the current pattern with an MPI_Allreduce.

ljcarlin:
In testing I noticed that `sc_io_read_at_all` doesn't correctly report read counts. The non-mpi version work fine, but the MPI version sometimes reports that the expected (nonzero) number of bytes have been read (even if the read location is the end of the file). I suspect this is an MPI bug rather than something on our end, but I'll talk to Tim about it tomorrow.

cburstedde:
> In testing I noticed that `sc_io_read_at_all` doesn't correctly report read counts. The non-mpi version work fine, but the MPI version sometimes reports that the expected (nonzero) number of bytes have been read (even if the read location is the end of the file). I suspect this is an MPI bug rather than something on our end, but I'll talk to Tim about it tomorrow.

Quite possible, @tim-griesbach has seen issues with the count before.  Sounds good!

tim-griesbach:
> Quite possible, @tim-griesbach has seen issues with the count before. Sounds good!

Louis and I had a quick look at the issue and it seems to be different from the count issues that I observed since I only observed transient count issues in the CI and these issues have not occurred for some time.

The issue that Louis found seems to reproducible with the MPI implementation that he uses. It may be an MPI implementation dependent issue.

jamabr:
Sorry for jumping in here, but maybe it helps. 
From my experience it is not sufficient to just attribute IO problems only to the MPI implementation. Not sure if you are aware of this. For instance OpenMPI has two IO backends included, the `ompio` backend and a quite old version of `romio321` (which is used in a newer version in MPICH). Since v2, OpenMPI switched the default backend to `ompio` so trying to use the `romio` backend is always a good option to see if the problem is within MPI. You can switch to the romio backend with ` export OMPI_MCA_io=^ompio`.

cburstedde:
Reading beyond EOF seems to be so rare that not all implementations test the respective code. Is there an alternative approach querying the file size first and not reading beyond?

cburstedde:
Hi @scivision we have a CI failure on Windows related to the zlib. Would you have an idea what's causing this?

ljcarlin:
> Reading beyond EOF seems to be so rare that not all implementations test the respective code. Is there an alternative approach querying the file size first and not reading beyond?

In this specific example the amount to read is read in at the start from the output of preprocessing, so it only actually matters if the user gives an incorrect file as input.

tim-griesbach:
> Reading beyond EOF seems to be so rare that not all implementations test the respective code. Is there an alternative approach querying the file size first and not reading beyond?

The MPI standard defines `MPI_File_get_size` to get the file size that could be check in the code before continuing. There is no portable, purely on the C-standard relying way to get file size without MPI I/O (cf. https://wiki.sei.cmu.edu/confluence/display/c/FIO19-C.+Do+not+use+fseek()+and+ftell()+to+compute+the+size+of+a+regular+file) but since we considering an issue that appears only with MPI I/O it may be sufficient to just use `MPI_File_get_size` as a workaround for the case of activated MPI I/O.

PR #289, opened 2024-03-01
Checksum comparison in wrap_adapt

hannesbrandt:
We add a checksum comparison in `wrap_adapt` to check whether coarsening and balancing cancel each other out.

One checksum is computed at the start of the function for the p4est of the input wrap. After refining, coarsening and balancing we compute another checksum, if the total amount of quadrants did not change. If the result p4est has the same checksum as the input p4est, we do not compute a new ghost and mesh for the wrap.

### DIFF
--- a/src/p4est_communication.c
+++ b/src/p4est_communication.c
@@ -942,12 +942,12 @@ p4est_comm_checksum (p4est_t * p4est, unsigned local_crc, size_t local_bytes)
 #ifdef P4EST_ENABLE_MPI
   int                 mpiret;
   int                 p;
-  uint64_t            send[2];
-  uint64_t           *gather;
+  long long           send[2];
+  long long          *gather;
 
-  send[0] = (uint64_t) local_crc;
-  send[1] = (uint64_t) local_bytes;
-  gather = P4EST_ALLOC (uint64_t, 2 * p4est->mpisize);
+  send[0] = (long long) local_crc;
+  send[1] = (long long) local_bytes;
+  gather = P4EST_ALLOC (long long, 2 * p4est->mpisize);
   mpiret = sc_MPI_Allgather (send, 2, sc_MPI_LONG_LONG_INT,
                              gather, 2, sc_MPI_LONG_LONG_INT, p4est->mpicomm);
   SC_CHECK_MPI (mpiret);
--- a/src/p4est_communication.c
+++ b/src/p4est_communication.c
@@ -947,24 +947,17 @@ p4est_comm_checksum (p4est_t * p4est, unsigned local_crc, size_t local_bytes)
 
   send[0] = (uint64_t) local_crc;
   send[1] = (uint64_t) local_bytes;
-  gather = NULL;
-  if (p4est->mpirank == 0) {
-    gather = P4EST_ALLOC (uint64_t, 2 * p4est->mpisize);
-  }
-  mpiret = sc_MPI_Gather (send, 2, sc_MPI_LONG_LONG_INT,
-                          gather, 2, sc_MPI_LONG_LONG_INT, 0, p4est->mpicomm);
+  gather = P4EST_ALLOC (uint64_t, 2 * p4est->mpisize);
+  mpiret = sc_MPI_Allgather (send, 2, sc_MPI_LONG_LONG_INT,
+                             gather, 2, sc_MPI_LONG_LONG_INT, p4est->mpicomm);
   SC_CHECK_MPI (mpiret);
 
-  if (p4est->mpirank == 0) {
-    for (p = 1; p < p4est->mpisize; ++p) {
-      crc = adler32_combine (crc, (uLong) gather[2 * p + 0],
-                             (z_off_t) gather[2 * p + 1]);
-    }
-    P4EST_FREE (gather);
+  crc = (uLong) gather[0];
+  for (p = 1; p < p4est->mpisize; ++p) {
+    crc = adler32_combine (crc, (uLong) gather[2 * p + 0],
+                           (z_off_t) gather[2 * p + 1]);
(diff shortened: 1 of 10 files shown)


cburstedde:
Thanks; I believe I am missing some updates in the p8est*.h files.

There is the suggestion to remove checksum_partition, can you recheck with @tim-griesbach if that function is (stil?) needed?

hannesbrandt:
> Once p4est_checksum is collective, would it make sense to go through our examples/tests and align them to the new behavior?

I have gone through the tests and adapted `test_io2.c` . The other tests and examples only compare checksums of different timesteps and did not need to be adapted.

Additionally, I switched the data types in `comm_checksum` from uint64_t to long long, since this is the format they are sent in.

src/p4est_communication.c
```diff
@@ -942,29 +942,22 @@ p4est_comm_checksum (p4est_t * p4est, unsigned local_crc, size_t local_bytes)
...
-  }
-  mpiret = sc_MPI_Gather (send, 2, sc_MPI_LONG_LONG_INT,
-                          gather, 2, sc_MPI_LONG_LONG_INT, 0, p4est->mpicomm);
+  long long           send[2];
+  long long          *gather;
+
+  send[0] = (long long) local_crc;
+  send[1] = (long long) local_bytes;
+  gather = P4EST_ALLOC (long long, 2 * p4est->mpisize);
+  mpiret = sc_MPI_Allgather (send, 2, sc_MPI_LONG_LONG_INT,
```
cburstedde, line 951: Would MPI_UNSIGNED_LONG_LONG be available? It may be a better fit.

src/p4est_wrap.c
```diff
@@ -695,16 +700,32 @@ p4est_wrap_adapt (p4est_wrap_t * pp)
...
                        pp->params.replace_fn);
-    pp->flags = P4EST_ALLOC_ZERO (uint8_t, p4est->local_num_quadrants);
 
-    pp->ghost_aux = p4est_ghost_new (p4est, pp->params.mesh_params.btype);
-    pp->mesh_aux =
-      p4est_mesh_new_params (p4est, pp->ghost_aux, &pp->params.mesh_params);
-    pp->match_aux = 1;
+    /* check if coarsening and balancing canceled out */
+    if (global_num_entry == p4est->global_num_quadrants) {
+      /* only compute another checksum, if the global quadrant count changed */
```
cburstedde, line 709: ... if it stayed the same?

src/p4est_wrap.c
```diff
@@ -695,16 +700,32 @@ p4est_wrap_adapt (p4est_wrap_t * pp)
...
+      checksum_exit = p4est_checksum (p4est);
+      changed = (checksum_entry != checksum_exit);      /* only relevant for rank 0 */
+      sc_MPI_Bcast (&changed, 1, sc_MPI_INT, 0, p4est->mpicomm);
```
cburstedde, line 712: Do we need the broadcast now that checksum is collective?

src/p4est_communication.c
hannesbrandt: It was introduced in the MPI 2.0 standard.

cburstedde: Ah ok then we don't use it.

cburstedde: We should probably document that currently any higher bits than 32 of the inputs to the function are ignored.

cburstedde: We do use MPI_UNSIGNED_LONG_LONG is libsc. Would we want to remove it, or should we configure test whether it exists?

cburstedde: While we're at it, ar there any other MPI 2  types in our sc_mpi code?

src/p4est_wrap.c
hannesbrandt: You are right, the broadcast is not needed anymore. I've updated this.

src/p4est_communication.c
cburstedde: Let's keep this MPI type for the function (possibly just with a doc update on too-large values). The question of removing MPI_UNSIGNED_LONG_LONG belongs into libsc.

hannesbrandt: That sounds good, should I create an issue for this in libsc?
Regarding the documentation, according to the C99 standard long long is at least 64 bits long. We might also change the parameters of `comm_checksum` to long long, although this might be inconsistent with the remaining input and output parameters.

cburstedde:
Would you think we're good to go on this one?

src/p4est_communication.c
cburstedde: Yes that would be nice, and checking which other types are > MPI 1.3.

Let us just document in the .h files of p4est that we use the adler32 checksum on the lower 32 bits of the input variables. By the way, does adler32 survive more than 32 bits total input length, iow, does it roll over cleanly once we exceed that range? But this may also be left for a different issue in p4est.

cburstedde: I'm leaning towards merging now and opening a p4est issue on 32 bit checksum issues.

hannesbrandt:
> Would you think we're good to go on this one?

I would like to address one last issue that I just noticed. Calling p4est_checksum aborts if zlib is not available. Should I adapt the code to only compute the checksum in wrap_adapt, if zlib is available?

cburstedde:
> > Would you think we're good to go on this one?
> 
> I would like to address one last issue that I just noticed. Calling p4est_checksum aborts if zlib is not available. Should I adapt the code to only compute the checksum in wrap_adapt, if zlib is available?

Yes that sounds good. If you could document this in the .h files?

cburstedde:
Thanks!

PR #290, opened 2024-03-01
Fix file cleanup

tim-griesbach:
# Fix file cleanup

Proposed changes: Fix the check for `file` being `NULL` in the function `p4est_file_error_cleanup` to prevent accessing `NULL` in the case of handling a failed call of the open function. In addition, a check abort message is updated.


cburstedde:
Thanks!

PR #291, opened 2024-03-03
Feature GMT distributed

ljcarlin:
Introduced a distributed mode to the GMT example, which can be enabled with the `-d` flag. In this mode each process knows only a fraction of the points and at each iteration points are sent to the relevant processes before refinement. Since points are known by multiple processes we avoid duplication by designating a single owner of each point as _responsible_ for propagating it in the next iteration.

Currently the sphere model is the only model set up for distributed mode. However, the communication code is written in such a way that other models could add support without writing their own communication code. 

To support distributed mode a model should (in setup) load a distinct subset of points on each process. Each process must set `model->M` as the number of points loaded on that process. A point must be represented by a struct of size `model->point_size` and points must be stored in the array `model->points`, which hence has byte-size `model->M * model->point_size`. The intersection function `model->intersect` should be written so that an input of `m` refers to the `m`th point stored in `model->points`.

The communication code uses `p4est_search_partition`  with the model-specified intersection function to determine which process domains each point intersects. Points are then communicated to the relevant processes, and the array `model->points` is updated, along with the local point count `model->M`.

### DIFF
--- a/example/gmt/Makefile.am
+++ b/example/gmt/Makefile.am
@@ -8,7 +8,8 @@ bin_PROGRAMS += example/gmt/p4est_gmt
 
 example_gmt_p4est_gmt_SOURCES = \
         example/gmt/gmt2.c \
-        example/gmt/gmt_models.c example/gmt/gmt_models.h
+        example/gmt/gmt_models.c example/gmt/gmt_models.h \
+        example/gmt/gmt_global.h
 
 bin_PROGRAMS += example/gmt/p4est_sphere_preprocessing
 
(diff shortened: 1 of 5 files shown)


cburstedde:
Thanks; is this a good moment to merge before continuing the communication abstraction?

ljcarlin:
> Thanks; is this a good moment to merge before continuing the communication abstraction?

I'll just make the change to stop processes messaging themselves and then it should be good to go.

ljcarlin:
Okay, that should be good to go @cburstedde.

I didn't change the context stuff we discussed, but I think it's better to think about that in the abstracted version.

PR #297, opened 2024-03-22
Feature transfer_search

ljcarlin:
This is an abstracted version of the algorithm I wrote for the GMT example.

`p4est_transfer_search` is a collective, point-to-point transfer for maintaining a distributed collection of points. After communication, points are stored (only) on the processes whose domains they intersect. 

Points can be instances of an arbitrary struct. Point-quadrant intersection is specified by a user supplied callback. A single point may intersect multiple process domains, and after communication will be known to each of these processes.

Each process is responsible for propagating a subset of the points it knows. Before communication, exactly one process should be responsible for propagating each point. The intersecting processes for each point are determined - by the responsible process - with `p4est_search_partition`. Points are then communicated to the relevant processes. Points known to a process before communication that do not intersect its domain are forgotten. The algorithm ensures that after communication exactly one process is responsible for the propagation of each point. This is the process with the lowest rank among processes intersecting the point.

I have rewritten my portion of the GMT example to use this abstracted version. This rewrite breaks the other parts of the example, due to intersection callbacks having a different type. The rewrite can be found here: https://github.com/ljcarlin/p4est/tree/feature-gmt

### DIFF
--- a/src/p4est_communication.h
+++ b/src/p4est_communication.h
@@ -586,16 +586,14 @@ void                p4est_transfer_end (p4est_transfer_context_t * tc);
 typedef int         (*p4est_intersect_t) (p4est_topidx_t which_tree,
                                           p4est_quadrant_t *quadrant,
                                           void *point, void *user);
-/* TODO: does this function need a p4est? Does it need a user pointer? */
-/* TODO: finish point callback, then fill out the rest of the template */
 
 /** The points being exchanged in \ref p4est_transfer_search. */
 typedef struct p4est_transfer_search {
   /* The points to exchange */
   sc_array_t *points;
 
-  /* Process is responsible for propagating the first num_resp points that it
-   * knows. The remaining points are additionally known to other processes
+  /* Each process is responsible for propagating the first num_resp points that
+   * it knows. The remaining points are additionally known to other processes
    * which are responsible for their propagation. */
   p4est_locidx_t num_resp;
 
@@ -639,7 +637,7 @@ void p4est_transfer_search_destroy (p4est_transfer_search_t *c);
  * communication.
  * 
  * \param [in] p4est     The forest we search with. Its user_pointer is passed
- *                      to the intersection callback.
+ *                       to the intersection callback.
  * \param [in,out] c     Points and propagation responsibilities.
  * \param [in] intersect Intersection callback.
  */
--- a/src/p4est_communication.h
+++ b/src/p4est_communication.h
@@ -603,7 +603,7 @@ typedef struct p4est_transfer_search {
 
 /** Destroy a \ref p4est_transfer_search_t structure
  * 
- * \param[in,out] c the structure to destroy
+ * \param [in,out] c the structure to destroy
  */
 void p4est_transfer_search_destroy (p4est_transfer_search_t *c);
 
@@ -638,10 +638,10 @@ void p4est_transfer_search_destroy (p4est_transfer_search_t *c);
  * subdivision if they modify the array of points between rounds of
  * communication.
  * 
- * \param[in] p4est     The forest we search with. Its user_pointer is passed
+ * \param [in] p4est     The forest we search with. Its user_pointer is passed
  *                      to the intersection callback.
- * \param[in,out] c     Points and propagation responsibilities.
- * \param[in] intersect Intersection callback.
+ * \param [in,out] c     Points and propagation responsibilities.
+ * \param [in] intersect Intersection callback.
  */
 int
 p4est_transfer_search (p4est_t *p4est, p4est_transfer_search_t *c, 
@@ -658,14 +658,16 @@ p4est_transfer_search (p4est_t *p4est, p4est_transfer_search_t *c,
  * \param [in] nmemb        Number of processors encoded in \a gfq (plus one).
  * \param [in] num_trees    Tree number must match the contents of \a gfq.
  * \param [in] user_pointer Passed to the intersection callback.
- * \param[in,out] c         Points and propagation responsibilities.
- * \param[in] intersect     Intersection callback.
+ * \param [in] mpicomm      Function is collective over the communicator.
+ * \param [in,out] c        Points and propagation responsibilities.
+ * \param [in] intersect    Intersection callback.
  */
 int
 p4est_transfer_search_gfx (const p4est_gloidx_t *gfq,
                             const p4est_quadrant_t *gfp,
                             int nmemb, p4est_topidx_t num_trees,
                             void *user_pointer,
+                            sc_MPI_Comm mpicomm,
                             p4est_transfer_search_t *c,
                             p4est_intersect_t intersect);
 
@@ -684,13 +686,15 @@ p4est_transfer_search_gfx (const p4est_gloidx_t *gfq,
  * \param [in] nmemb        Number of processors encoded in \a gfq (plus one).
  * \param [in] num_trees    Tree number must match the contents of \a gfq.
  * \param [in] user_pointer Passed to the intersection callback.
- * \param[in,out] c         Points and propagation responsibilities.
- * \param[in] intersect     Intersection callback.
+ * \param [in] sc_MPI_Comm  Function is collective over the communicator.
+ * \param [in,out] c        Points and propagation responsibilities.
+ * \param [in] intersect    Intersection callback.
  */
 int
 p4est_transfer_search_gfp (const p4est_quadrant_t *gfp, int nmemb,
                             p4est_topidx_t num_trees,
                             void *user_pointer,
+                            sc_MPI_Comm mpicomm,
                             p4est_transfer_search_t *c,
                             p4est_intersect_t intersect);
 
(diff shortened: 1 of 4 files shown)


src/p4est_communication.h
```diff
@@ -575,6 +575,141 @@ void                p4est_transfer_items_end (p4est_transfer_context_t * tc);
...
+   * which are responsible for their propagation. */
+  p4est_locidx_t num_resp;
+
+} p4est_transfer_search_t;
+
+/** Destroy a \ref p4est_transfer_search_t structure
+ * 
+ * \param [in,out] c the structure to destroy
+ */
+void p4est_transfer_search_destroy (p4est_transfer_search_t *c);
```
cburstedde, line 606: We may remove this function and instead provide a full documentation of the transfer_search_t on input and output of the transfer_search functions.

ljcarlin:
@cburstedde I added more documentation and a `\file` so that `p4est_communication.h` is parsed by doxygen. If it was intentional not to have that then I can remove it again.

I also noticed that doxygen mistakes `\ref p4est_transfer_search` as referring to the structure `p4est_transfer_search` behind the type `p4est_transfer_search_t` rather than the function `p4est_transfer_search`. Basically when you typedef and define a struct at the same time then doxygen interprets the comment as referring to the struct rather than the type. I think this happens elsewhere in p4est too.

cburstedde:
> @cburstedde I added more documentation and a `\file` so that `p4est_communication.h` is parsed by doxygen. If it was intentional not to have that then I can remove it again.

Cool; thanks for the effort!

> I also noticed that doxygen mistakes `\ref p4est_transfer_search` as referring to the structure `p4est_transfer_search` behind the type `p4est_transfer_search_t` rather than the function `p4est_transfer_search`. Basically when you typedef and define a struct at the same time then doxygen interprets the comment as referring to the struct rather than the type. I think this happens elsewhere in p4est too. 

This is unfortunate; thanks for figuring this out.  So would we want to
rename every struct behind a typedef by adding _s?  That would affect quite
a bit of the code.

ljcarlin:
> This is unfortunate; thanks for figuring this out. So would we want to rename every struct behind a typedef by adding _s? That would affect quite a bit of the code.

Yes, it seems quite unfortunate.

Another approach is to declare the struct first (without documentation) and then document the _t typedef. I think this makes the code a bit less readable though, and it requires just as much code change.

src/p4est_communication.h
```diff
@@ -575,6 +582,184 @@ void                p4est_transfer_items_end (p4est_transfer_context_t * tc);
...
+ *                          this function has been called.
+ * \param [in] intersect    Intersection callback.
+ * \param [in] save_unowned If true then points that would be unowned are
+ *                          maintained by their propagating process
+ * \return 0 if transfer was successful.
+ */
+int
+p4est_transfer_search_gfx (const p4est_gloidx_t *gfq,
+                            const p4est_quadrant_t *gfp,
+                            int nmemb, p4est_topidx_t num_trees,
```
cburstedde, line 722: Thanks! I'm wondering: can nmemb be inferred by the size of the communicator, and can num_trees be read from the last element of gfp? This would save two arguments.

src/p8est_communication.h
```diff
@@ -564,6 +564,181 @@ void                p8est_transfer_items_end (p8est_transfer_context_t * tc);
...
+ * time. Care should be taken when adding or deleting points, and when
+ * modifying the order of \a points, to ensure that \a num_resp is updated
+ * and that the first \a num_resp points are still the points that the
+ * process should propagate. 
+ * 
+ * During the transfer of points in \ref p8est_transfer_search the \a points
+ * array is destroyed and reallocated. Thus users should not maintain pointers
+ * to it or its contents.
+ */
+typedef struct p8est_transfer_search
```
cburstedde, line 621: I'd be fine suffixing the struct name with _s to disambiguate for doxygen.

src/p4est_communication.h
cburstedde: On the other hand, we don't do it like this in the transfer functions. We should use the information passed for very strict assertion checking to maximize robustness.

src/p4est_communication.h
```diff
@@ -575,6 +582,184 @@ void                p4est_transfer_items_end (p4est_transfer_context_t * tc);
...
+ * partition, rather than an explicit p4est. The partition can be that of any
+ * p4est, not necessarily known to the caller.
+ * 
+ * This function is collective.
+ * 
+ * \param [in] gfq          Partition offsets to traverse.  Length \a nmemb + 1.
+ * \param [in] gfp          Partition position to traverse.  Length \a nmemb + 1.
+ * \param [in] nmemb        Number of processors encoded in \a gfq (plus one).
+ * \param [in] num_trees    Tree number must match the contents of \a gfq.
+ * \param [in] user_pointer Passed to the intersection callback.
```
cburstedde, line 708: Can we be more explicit on the user_pointer being part of a dummy p4est, and which elements of the dummy p4est are meaningfully populated?

cburstedde:
> @cburstedde I added more documentation and a `\file` so that `p4est_communication.h` is parsed by doxygen. If it was intentional not to have that then I can remove it again.

This is very helpful; thanks! We're behind on some documentation.

> I also noticed that doxygen mistakes `\ref p4est_transfer_search` as referring to the structure `p4est_transfer_search` behind the type `p4est_transfer_search_t` rather than the function `p4est_transfer_search`. Basically when you typedef and define a struct at the same time then doxygen interprets the comment as referring to the struct rather than the type. I think this happens elsewhere in p4est too.

Outside of this PR, which other struct names may be in need of a _s suffix?

src/p4est_communication.h
cburstedde: (This may be a duplicate comment.) Otoh, we're keeping the same redundancy in the search functions. So maybe we stick to this convention and make very sure all parameters are consistent with each other.

src/p4est_communication.h
cburstedde: Maybe we should define a central p4est_init_dummy function to standardize dummy use wherever it occurs.

cburstedde: > Maybe we should define a central p4est_init_dummy function to standardize dummy use wherever it occurs.

This may be better done in a separate PR though.

src/p8est_communication.h
cburstedde: I've opted to rename the struct to p4est_points_context. Problem solved -- for now.

cburstedde:
I'm now thinking about the unowned points. By definition they are part of the propagated portion of the points array (?). Should we indicate which of the propagated are unowned? We may put these into a contiguous range right in the beginning and add a num_unowned member?

ljcarlin:
> Outside of this PR, which other struct names may be in need of a _s suffix?

I noticed you fixed this just by changing the name of the struct. This is probably a better solution. I don't think the issue actually appears very frequently, I had a quick look and couldn't see it anywhere else.

ljcarlin:
> I'm now thinking about the unowned points. By definition they are part of the propagated portion of the points array (?). Should we indicate which of the propagated are unowned? We may put these into a contiguous range right in the beginning and add a num_unowned member?

Sure, that sounds reasonable. As is, a search based refinement would be refining with the irrelevant unowned points, which is a bit silly.

cburstedde:
> > Outside of this PR, which other struct names may be in need of a _s suffix?
> 
> I noticed you fixed this just by changing the name of the struct. This is probably a better solution. I don't think the issue actually appears very frequently, I had a quick look and couldn't see it anywhere else.

Thanks!

cburstedde:
> > I'm now thinking about the unowned points. By definition they are part of the propagated portion of the points array (?). Should we indicate which of the propagated are unowned? We may put these into a contiguous range right in the beginning and add a num_unowned member?
> 
> Sure, that sounds reasonable. As is, a search based refinement would be refining with the irrelevant unowned points, which is a bit silly.

Cool, thanks for the update!  It seems with this, the unowned feature is now
fully functional as seen from the outside (?).

I have one minor suggestion:  We might define a dummy p4est to always have
valid mpicomm, mpisize, and mpirank members in addition to the user_pointer
possibly being set.  The rest of the bytes of the p4est struct shall be
initialized to zero.  This would allow us to cut a few communicator-related
arguments passed to static functions.  Similarly, we may not pass point_size
in that one instance where the internal structure is passed already.

About the error handling: with 'errsend = errsend || ...' the compiler omits
the execution of the second expression if the first is already true.  It
seems that this is intentional, just checking that per-rank error conditions
don't hang the parallel execution of the algorithm.

And, lastly, may I ask to run `./p4estindent` on the updated files?  Beware
however that some 'features' of that script require manual override, such as
obvious nonsense and function arguments: use `(type_t *var)`, not `* var`.

cburstedde:
What we might also do is adding a third variable to the context, num_total. This is redundant but maybe a nice feature for consistency.

cburstedde:
Ok, let's not to the dummy p4est since your code aligns perfectly with the search_gf* routines.

Just remaining to look into the errsend question and indenting the .c file (I've done it for the .h).

PR #298, opened 2024-03-28
Ensure C standard compliant `fseek` usage

tim-griesbach:
# Ensure C standard compliant `fseek` usage

According to the C standard (cf. https://wiki.sei.cmu.edu/confluence/display/c/FIO19-C.+Do+not+use+fseek()+and+ftell()+to+compute+the+size+of+a+regular+file), it can not be assumed that seeking to the end of a binary file works in general. However, this was done in `p4est_save_ext`. In concrete terms, the C standard states that
```
Setting the file position indicator to end-of-file, as with fseek(file, 0, SEEK_END), has undefined behavior for a binary stream (because of possible trailing null characters) or for any stream with state-dependent encoding that does not assuredly end in the initial shift state.
```

Proposed changes: This PR ensures that it is not necessary to seek to the end of a binary file by not closing the file in `p4est_save_ext` after the connectivity was written to disk.


src/p4est_connectivity.h
```diff
@@ -446,6 +446,18 @@ sc_array_t         *p4est_connectivity_deflate (p4est_connectivity_t * conn,
                                                 p4est_connectivity_encode_t
                                                 code);
 
+/** Save a connectivity structure to disk and not close the file.
+ * \param [in] filename         Name of the file to write.
+ * \param [in] connectivity     Valid connectivity structure.
+ * \param [in,out] sink_out     A pointer to pointer to set the sc_io_sink_t.
+ * \return                      Returns 0 on success, nonzero on file error.
+ */
+int                 p4est_connectivity_save_preserve (const char *filename,
```
cburstedde, line 455: Wondering: can we use p4est_connectivity_sink and then write the forest using sc_io_sink functions, to not be using sink->file? One may also ask whether we should not introduce a function p4est_sink that will share a lot of code with p4est_save.

tim-griesbach: I can replace the `sink->file` occurrences with sc_io_sink function calls. 

However, the `save_ext` is currently using `fseek` on all ranks > 0 with respect to I/O operations that are started with `fopen` and not with `sc_io_sink_new`. To my understanding, seeking is not supported by the sc_io_sink functions.  Hence, I think that without such a functionality for sc_io_sink we can not write the forest only with sc_io_sink functions but we still can avoid `sink->file`.

tim-griesbach:
I removed the occurrences of `sink->file` by using `p4est_connectivity_sink` in `p4est_save_ext`.

tim-griesbach:
sc_test_version failed in one CI test. This should be unrelated to the changes in this PR and transient.

PR #300, opened 2024-04-01
Cmake examples

mkirilin:
This PR simplifies and unifies examples/CMakeLists.txt file.
In addition, it make examples build by default.

### DIFF
--- a/example/CMakeLists.txt
+++ b/example/CMakeLists.txt
@@ -44,14 +44,14 @@ set_target_properties(${name}
 endfunction(p8est_example)
 
 
-function(p4est_example_test name dir)
+function(p4est_p8est_example name dir)
 
 p4est_example(${name}2 ${dir}/${name}2.c ${dir})
 
 if(enable_p8est)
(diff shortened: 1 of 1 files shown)


cburstedde:
Thanks! is this as much as possible analogous to the libsc code for the examples?

mkirilin:
It is a good balance between simplicity and similarity with the libsc, considering the peculiarities  of p4est (e.g. 2 or 3 dimensions, etc.)

cburstedde:
> It is a good balance between simplicity and similarity with the libsc, considering the peculiarities  of p4est (e.g. 2 or 3 dimensions, etc.)

Cool!  CMake has its own way of defining enable p4est, p6est, p8est.  Do the
examples compile in all combinations?

mkirilin:
> Cool!  CMake has its own way of defining enable p4est, p6est, p8est.  Do the
> examples compile in all combinations?

Yes, I've tried all the 4 combinations of p6est / p8est and it works well.

PR #301, opened 2024-04-03
Check if libsc is already a known target.

jmark:
# Check if libsc is already a known target.

Proposed changes:

This PR adds a check if libsc is already a known target in the CMake build system. If
this is the case, registering libsc is skipped.

This is useful for cases when p4est is included as dependency by other software projects, e.g. t8code,
which in turn also depends on libsc. If this check would not be there CMake registers libsc twice which
causes conflicts.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,7 +20,7 @@ message(STATUS "p4est ${PROJECT_VERSION} "
 
 # Skip `libsc` if already registered. This is useful when `p4est` is added as a
 # dependency by another software project which in turn already registered `libsc`.
-if(NOT TARGET sc)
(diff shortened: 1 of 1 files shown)


cburstedde:
Thanks; well spotted. @scivision, would you think this works given other hierarchical projects?

jmark:
> I would use
> 
> ```cmake
> if(NOT TARGET SC::SC)
> ```
> 
> as both targets have the same scope. The "SC::SC" is a bit easier to grok when finding where something came from.

@scivision I applied your suggestion. Thanks for the advice!

cburstedde:
Thanks so much @jmark and @scivision!

PR #184, opened 2023-03-20
Add my license statement.

bangerth:

Issue #182, opened 2023-03-19
Fixing a typo in example/CMakeLists.txt and suggesting two related minor changes


pkestene:
**Description**

This issue is about cmake build system in branch prev3-develop.

1. If think the following line
https://github.com/cburstedde/p4est/blob/prev3-develop/example/CMakeLists.txt#L11
should be changed into
```cmake
list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/../cmake)
```
since subfolder `Modules` doesn't exists.

2. This is about a minor change, I'd like to suggest:

When trying to build p4est examples using cmake, for example on a machine where p4est/sc is already installed by the system (e.g. using deb package on Ubuntu machine), I found that the CMakeCache.txt is kind of polluted because variable SC_CONFIG_H is set to  `/usr/include/sc_config.h` eventhough I used CMAKE_PREFIX_PATH to the location where I've just built the latest version of p4est (branch prev3-develop). So this information (SC_CONFIG_PATH set to /usr/include/sc_config.h) gives the wrong impression that cmake detected the system installation.

Actually this behavior is related to the way example/CMakeLists.txt is trying to detect SC et P4EST :
```cmake
find_package(SC REQUIRED)
find_package(P4EST REQUIRED)
```
The macro `find_package` will ALWAYS first use the MODULE mode to detect SC and P4EST (that means use FindSC.cmake and FindP4EST.cmake), and if the detection fails, it will then use the CONFIG mode (that means use SCConfig.cmake and P4ESTConfig.cmake).

So no matter what CMAKE_PREFIX_PATH is, find_package(SC REQUIRED), will always set variable SC_CONFIG_H to /usr/include/sc_config.h on a system which has p4est/sc install in system path, during the MODULE pass which will fail, and succeed later during the CONFIG pass of find_package(SC REQUIRED).

So finally, I suggest the following change
```cmake
find_package(SC CONFIG)
if(NOT SC_FOUND)
  find_package(SC REQUIRED)
endif()

find_package(P4EST CONFIG)
if(NOT P4EST_FOUND)
  find_package(P4EST REQUIRED)
endif()
```
thus, we always first try the CONFIG mode, if it fails, only then we required the MODULE mode to succeed. 
[...]

PR #303, opened 2024-05-10
CMake updates like libsc

scivision:
Like libsc #178, #182, and #184

* no default install prefix as best practice
* simply CI YaML w/o compiler versions like libsc

### DIFF
--- a/.github/workflows/composite-unix/action.yml
+++ b/.github/workflows/composite-unix/action.yml
@@ -1,0 +1,38 @@
+runs:
+
+  using: 'composite'
+
+  steps:
+  - name: CMake configure
+    shell: bash
+    run: >-
+      cmake --preset default
+      -DBUILD_SHARED_LIBS:BOOL=${{ matrix.shared }}
+
+  - name: CMake build
+    shell: bash
+    run: cmake --build --preset default
+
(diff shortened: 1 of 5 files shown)


cburstedde:
Thanks; the corresponding libsc PR had some ignore entries in the non-Cmake CI runners, would this be the same logic here?

.github/workflows/composite-unix/action.yml
```diff
@@ -0,0 +1,38 @@
+runs:
+
+  using: 'composite'
+
+  steps:
+  - name: CMake configure
+    shell: bash
+    run: >-
+      cmake --preset default
+      -DBUILD_SHARED_LIBS:BOOL=${{ matrix.shared }}
```
cburstedde, line 10: Wondering: Will we add the P4EST_ prefix to these variables?

cburstedde:
Cool, let's get the last of the sc PRs merged, verify that the CI is good for CMake, and then align this one analogously.

cburstedde:
Thanks for your patience, I've updated the p4est develop branch with the latest libsc. The conflicts may go away easily?

cburstedde:
Cool thanks!

PR #306, opened 2024-05-17
CMake option to use a system installed sc

sandro-elsweijer:
# Add CMake option to use a system installed SC 

Proposed changes:
For our new t8code CMake CI we need the option to build p4est with a system installed SC version.
This option already exists in p4ests autotools build system but does not exist in the CMake build system.
This PR adds this option. The standard behavior of p4ests CMake build system is not altered since the standard policy is to use the SC which ships with p4est.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,19 +18,17 @@ message(STATUS "p4est ${PROJECT_VERSION} "
 
 # --- external libs
 
-# Option to use an already installed SC library.
-if ( P4EST_USE_SYSTEM_SC )
-    find_package( SC REQUIRED PATHS /path/to/system/sc )
-else()
-    add_subdirectory( ${CMAKE_CURRENT_LIST_DIR}/sc )
-endif()
-
 # Skip `libsc` if already registered. This is useful when `p4est` is added as a
 # dependency by another software project which in turn already registered `libsc`.
 if(NOT TARGET SC::SC)
-  include(cmake/GitSubmodule.cmake)
-  git_submodule("${PROJECT_SOURCE_DIR}/sc")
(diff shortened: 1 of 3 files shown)


sandro-elsweijer:
Whoopsie, I did not see the other compile option and merged them now.

cburstedde:
Thanks so much! We're in the middle of merging a few CMake updates in libsc and will then align p4est to work the same way as much as possible. Let's keep an eye on these merges to make sure your code works cleanly with those.

CMakeLists.txt
```diff
@@ -21,9 +21,14 @@ message(STATUS "p4est ${PROJECT_VERSION} "
 # Skip `libsc` if already registered. This is useful when `p4est` is added as a
 # dependency by another software project which in turn already registered `libsc`.
 if(NOT TARGET SC::SC)
-  include(cmake/GitSubmodule.cmake)
-  git_submodule("${PROJECT_SOURCE_DIR}/sc")
-  add_subdirectory(sc)
+  # Option to use an already installed SC library.
+  if ( P4EST_USE_SYSTEM_SC )
+      find_package( SC REQUIRED PATHS /path/to/system/sc )
```
cburstedde, line 26: I'm not too familiar with CMake, are we really using `/path/to/system/sc` verbatim?

sandro-elsweijer: Done

cburstedde:
Thanks again! I'd be hoping that this runs fine on top of the latest develop branch.

sandro-elsweijer:
@cburstedde Done, at least locally all tests pass with the libsc shipped by p4est and also with a locally installed libsc (:

PR #310, opened 2024-07-01
Newly adapted quadrants in wrap

hannesbrandt:
This PR adds two arrays `newly_refined` and `newly_coarsened` to the `p4est_wrap_t`. The arrays store the local quadrant indices of quadrants that were refined or coarsened during the most recent call to `p4est_wrap_adapt`. For refinement, only the index of the quadrant with child id 0 is stored. The arrays are not updated during other function calls like `p4est_wrap_partition`. So, they always index into the local quadrants of the p4est as it was right after adaptation.

The arrays are only initalized and set, if the new wrap parameter `store_adapted` is set to true. Furthermore, we assert that the wrap's p4est is balanced before the adaptation, which ensures a single level of refinement.

### DIFF
--- a/src/p4est_wrap.c
+++ b/src/p4est_wrap.c
@@ -865,11 +865,9 @@ p4est_wrap_adapt (p4est_wrap_t * pp)
           qz++;
         }
       }
-      P4EST_ASSERT (zz == tquadrants->elem_count
-                    || pp->params.partition_for_coarsening == 0);
+      P4EST_ASSERT (zz == tquadrants->elem_count);
     }
-    P4EST_ASSERT (qz == quad_levels->elem_count
-                  || pp->params.partition_for_coarsening == 0);
+    P4EST_ASSERT (qz == quad_levels->elem_count);
   }
   sc_array_destroy (quad_levels);
 
--- a/src/p4est_wrap.c
+++ b/src/p4est_wrap.c
@@ -255,11 +255,6 @@ p4est_wrap_new_p4est_params (p4est_t * p4est, p4est_wrap_params_t * params)
       p4est_mesh_new_params (pp->p4est, pp->ghost, &pp->params.mesh_params);
   }
 
-  if (pp->params.store_adapted) {
-    pp->newly_refined = sc_array_new (sizeof (p4est_locidx_t));
-    pp->newly_coarsened = sc_array_new (sizeof (p4est_locidx_t));
-  }
-
   /* reset the data size since changing the p4est_wrap will affect p.user_int */
   p4est_reset_data (pp->p4est, 0, NULL, NULL);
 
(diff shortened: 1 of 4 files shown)


cburstedde:
Cool; would we want to delete the arrays once they no longer match the p4est (e. g. due to partition)?

hannesbrandt:
I have changed the allocation of the arrays. `newly_refined` and `newly_coarsened` are now allocated in wrap_adapt, instead of in wrap_new. As a consequence, `p4est_wrap_set_store_adapted` would now merely change the value of the corresponding wrap parameter, so I removed it. We could remove `p4est_wrap_set_partitioning` as well, since it just changes the partition_for_coarsening parameter.

src/p4est_wrap.h
```diff
@@ -102,6 +107,17 @@ typedef struct p4est_wrap
...
+  /* If \a params.store_adapted is true, these arrays store the indices of the
+   * quadrants refined and coarsened during the most recent call to
+   * \ref p4est_wrap_adapt. The wrap's \a p4est has to be balanced when entering
+   * the adaptation, to avoid multi-level refinement.
+   * The arrays are allocated during the first call of \ref p4est_wrap_adapt.
+   * At every time the arrays index into the local quadrants of the p4est as it
+   * was directly after completion of \ref p4est_wrap_adapt. So, they are not
+   * updated in \ref p4est_wrap_partition. Newly_refined only stores newly
+   * refined quadrants with child id 0. */
+  sc_array_t         *newly_refined, *newly_coarsened;
```
cburstedde, line 119: Can we have doxygen format documentation for these two variables?

hannesbrandt: Yes, I have added a short, doxygen-format comment for each of them.

cburstedde: Cool!

src/p4est_wrap.c
```diff
@@ -319,6 +320,26 @@ p4est_wrap_new_copy (p4est_wrap_t * source, size_t data_size,
     p4est_reset_data (pp->p4est, data_size, NULL, NULL);
   }
 
+  /* copy newly_adapted arrays if set */
+  P4EST_ASSERT (source->params.store_adapted ||
```
cburstedde, line 324: Is it not possible that the arrays are non-NULL and afterwards the parameter store_adapted is changed from 1 to 0? On the other hand, parameters are no longer changed during the life-time of the wrap object officially, but unofficial uses may exist.

cburstedde: We may adopt the following principle: When working with the stored arrays, we query them for being NULL or not and generally ignore the parameter. Only when deciding to re-allocate them after adaptation do we query the parameter. Would that make sense?

Then the only assertion that we may need is (newly_refined == NULL) == (newly_coarsened == NULL).

hannesbrandt: This seems like the most consistent convention, I have updated the code accordingly.

cburstedde:
Looks good; thanks!

PR #315, opened 2024-08-01
Add tree offsets to newly adapted quadrants in wrap

hannesbrandt:
This PR adds the tree offsets to the indices of recently adapted quadrants, which are stored in the `newly_refined` and `newly_coarsened` arrays introduced in https://github.com/cburstedde/p4est/pull/310, thereby ensuring a unique index for every locally refined quadrant.


cburstedde:
Thanks!

PR #316, opened 2024-09-13
CMake P4EST_BUILD_EXAMPLES flag

dutkalex:
# CMake P4EST_BUILD_EXAMPLES flag
Just adds this flag to give the opportunity to downstream projects to disable this and save on build times.


cburstedde:
Thanks! We had previously set examples to be always built, but it's fine to add this option as you suggest with default ON. While we're at it:

 - Can we make an analogous change in libsc, double-checking that the CMake example logic looks the exact same in both libraries?
 - I would not build the tests by default, only on make check, as with autoconf. Is this possible with CMake (and again, for both p4est and sc)?

dutkalex:
> * Can we make an analogous change in libsc, double-checking that the CMake example logic looks the exact same in both libraries?

Yes that seems reasonable. I can open a quick PR on libsc too.

> * I would not build the tests by default, only on make check, as with autoconf. Is this possible with CMake (and again, for both p4est and sc)?

That I do not know. Maybe it can be done, but no obvious solution comes to my mind right now...

cburstedde:
> > * Can we make an analogous change in libsc, double-checking that the CMake example logic looks the exact same in both libraries?
> 
> Yes that seems reasonable. I can open a quick PR on libsc too.

Sounds great! There seems to be just some small glitch with the CI. And please add your author file to the docs, and a line to the doc/release_notes.txt

> > * I would not build the tests by default, only on make check, as with autoconf. Is this possible with CMake (and again, for both p4est and sc)?
> 
> That I do not know. Maybe it can be done, but no obvious solution comes to my mind right now...

Thanks; then we let this remain as is for now.

cburstedde:
Same here, presumably, please submit against develop.

dutkalex:
Done @cburstedde

cburstedde:
Pinging @tim-griesbach for the current CI issues accumulating. Can we streamline the CI by taking out specific version numbers etc., and checking which packages we need on Darwin?

cburstedde:
Cool thanks!

PR #318, opened 2024-10-11
Add geometry for pillow connectivity to map the sphere

pkestene:
Add missing geometry for this connectivity.

The transformation from logical tree-local coordinates to physical coordinates to map the sphere are based upon:

"Logically rectangular finite volume methods with adaptive refinement on the sphere", Berger et al.

https://doi.org/10.1098/rsta.2009.0168

Issue #304, opened 2024-05-16
Ubuntu 24.04 with libmpich-dev fails test_loadsave2


scivision:
The "test_loadsave2" test passes on Ubuntu 24.04 with OpenMPI, and on other OS (Ubuntu 22.04, macOS, etc) regardless of OpenMPI or MPICH.

However, discovered in #303 and confirmed on a laptop with Ubuntu 24.04 is that "test_loadsave2" [fails with MPICH and GCC-12, GCC-13, or GCC-14](https://github.com/scivision/p4est/actions/runs/9114270509/job/25058539340)

Zlib and MPICH were enabled/used.

Wondering if this is just a flaky test or a code update is needed?

```
$ mpiexec -n 2 test/test_loadsave2

[libsc] This is libsc 2.8.6.999
[libsc] CPP                      /usr/bin/mpicc -E
[libsc] CPPFLAGS
[libsc] CC                       /usr/bin/mpicc
[libsc] CFLAGS                    -flto=auto;-ffat-lto-objects
[libsc] LDFLAGS                  -Wl,-Bsymbolic-functions
[libsc] LIBS                     /usr/lib/x86_64-linux-gnu/libz.so m
[p4est] This is p4est 0.0.0
[p4est] CPP                      /usr/bin/mpicc -E
[p4est] CPPFLAGS
[p4est] CC                       /usr/bin/mpicc
[p4est] CFLAGS                    -flto=auto;-ffat-lto-objects
[p4est] LDFLAGS                  -Wl,-Bsymbolic-functions
[p4est] LIBS                        m
[p4est] Using file names p4est.p4c and p4est.p4p
[p4est] Into p4est_new with min quadrants 0 level 0 uniform 0
[p4est] New p4est with 6 trees on 1 processors
[p4est] Initial level 0 potential global quadrants 6 per tree 1
[p4est] Done p4est_new with 6 total quadrants
[p4est] Into p4est_refine with 6 total quadrants, allowed level 29
[p4est] Done p4est_refine with 116163 total quadrants
[p4est] Into p4est_inflate
[p4est] Done p4est_inflate
[libsc] This is libsc 2.8.6.999
[libsc] CPP                      /usr/bin/mpicc -E
[libsc] CPPFLAGS
[libsc] CC                       /usr/bin/mpicc
[libsc] CFLAGS                    -flto=auto;-ffat-lto-objects
[libsc] LDFLAGS                  -Wl,-Bsymbolic-functions
[...]

PR #321, opened 2024-10-16
Add geometry for 2d pillow connectivity to map the sphere and new pillow3d to map a spherical shell

pkestene:
- add missing geometry for pillow connectivity for mapping a spherical shell
- add new 2d geometry `pillow_disk` for mapping a 2d disk from unitsquare connectivity
- add new 3d connectivity `pillow3d` and associated geometry to map a spherical shell

The transformation used in pillow to map the sphere are based upon:
"Logically rectangular finite volume methods with adaptive refinement on the sphere", Berger et al.
https://doi.org/10.1098/rsta.2009.0168

The transformation used in pillow_disk to map the disk are based upon:
"Logically rectangular grids and finite volume methods for PDEs in circular and spherical domains", Calhoun et al, SIAM Review, volume 50, Issue 4, January 2008.
https://doi.org/10.1137/060664094

This PR is a also a fix of #318 (dropped because pillow3d connectivity was erroneous).


example/simple/simple2.c
```diff
@@ -408,7 +426,30 @@ main (int argc, char **argv)
...
+  else if (config == P4EST_CONFIG_PILLOW_DISK) {
+    double               R = 1.0;        /* disk radius default value */
+    int                  iconfig;
+    pillow_disk_config_t pconfig = FIG3A;
+
+    if (argc >= 4)
+      R = atof (argv[3]);
+    if (argc >= 5) {
+      iconfig = atoi (argv[4]);
+      if (iconfig >=0 && iconfig <=3) {
```
cburstedde, line 446: Would it be safer to check the range using the names FIG3A to FIG3D?

cburstedde:
Cool, thanks! I'm thinking about unifying the naming. What about in 3D: just call it pillow instead of pillow3d. For symmetry with 2D, we may want to add in 3D a 1-tree pillow_sphere geometry if you like the idea.

Details: running indent would be great (ignoring its changes wrt. whitespace in argument lists (use: type *variable).

pkestene:
> Cool, thanks! I'm thinking about unifying the naming. What about in 3D: just call it pillow instead of pillow3d. For symmetry with 2D, we may want to add in 3D a 1-tree pillow_sphere geometry if you like the idea.
> 
> Details: running p4estindent would be great (ignoring its changes wrt. whitespace in argument lists (use: type *variable)).

Ok, I'll modify the code to take into account your remarks, and come back shortly.

pkestene:
Hi @cburstedde 
I've updated this branch to address your comments and rebased on latest develop.

pkestene:
Hi @cburstedde

would it be possible to re-run ci, since the failure are related to the transient problem of test_loadsave2 (issue #304) but apparently not with the proposed modifications ?

cburstedde:
Sorry for the wait, we had to work around some CI issues. I've now merged the latest develop into this PR.

cburstedde:
Thanks so much! Those are really some fun and versatile geometries.

Issue #325, opened 2024-11-18
P4est version is set to 0.0.0 if built from tarball via CMake


jmark:
**Description**

When p4est sources are packaged via a CMake tarball the version is set to 0.0.0.
The result in `include/p4est_config.h` reads
```
[...]

/* Define to the full name and version of this package. */
#define P4EST_PACKAGE_STRING "p4est 0.0.0"

/* Define to the one symbol short name of this package. */
#define P4EST_PACKAGE_TARNAME "p4est"

/* Define to the home page for this package. */
#define P4EST_PACKAGE_URL ""

/* Define to the version of this package. */
#define P4EST_PACKAGE_VERSION "0.0.0"

/* Version number of package */
#define P4EST_VERSION "0.0.0"

/* Package major version */
#define P4EST_VERSION_MAJOR 0

/* Package minor version */
#define P4EST_VERSION_MINOR 0

/* Package point version */
#define P4EST_VERSION_POINT 0
```

**To Reproduce**

Configure p4est with CMake and then create a tarball distribution via
```
$ make package_source
```

Then extract the tarball again and configure it via CMake:
```
tar xvf ${P4EST_BUILD_DIR}/package/p4est-.tar.gz
cmake
cat include/p4est_config.h
```

The naming of the tarball is also odd: `p4est-.tar.gz`.

**Additional information**

This is related to this issue: https://github.com/DLR-AMR/t8code/issues/1294

In t8code we had the same issue. The problem is that the tarball is not a Git repo anymore
and `git-gen version` is not available. Thus, the version info has to be retrieved another way.

t8code writes out a `version.txt` file and reads it in case of a tarball distribution.
See https://github.com/DLR-AMR/t8code/blob/main/cmake/GitProjectVersion.cmake
and https://github.com/DLR-AMR/t8code/blob/f8b7becc597975de218d40fb7804ec9723164e04/cmake/CPackConfig.cmake#L24.

If you find a more elegant solution, I'm all ears!

jmark:
The issues applies to libsc as well.

tim-griesbach:
Thank you for your report!

I created a [draft PR](https://github.com/cburstedde/p4est/pull/326) that creates a file similar to your `version.txt`. However, I create a slightly different file `.tarball-version` to be closer to the tarball creation with Autotools in p4est. The `git-version-gen` script can then parse `.tarball-version` as it does it for the Autotools workflow in p4est.

[...]

PR #326, opened 2024-11-18
Fix tarball versioning

tim-griesbach:
# Fix tarball versioning

This PR addresses among others the issue of tarball naming in the CMake build system. Previously, an uninitialized variable, `git_version`, was used for the version in the package name, leading to incorrect tarball names.

Moreover,  this PR introduces a new CMake target `GenerateVersionFile` to generate the file `.tarball-version` analogous to the file used in the Autotools-based creation of a tarball in p4est.

Then the file `.tarball-version` is used by the script `git-version-gen` to get the version in the case `.git` is not present, as it is the case for creating a tarball. This ensures a correctly populated `p4est_config.h` after configuring in a tarball directory.

Currently, the CMake target `GenerateVersionFile` is only build if some arbitrary build command is executed. In particular calling only `make package_source` does not suffice to trigger `.tarball-version`'s creation.

### DIFF
--- a/cmake/install.cmake
+++ b/cmake/install.cmake
@@ -49,7 +49,7 @@ add_custom_command(
     DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/CMakeLists.txt
   )
 
-add_custom_target(P4EST_GenerateVersionFile ALL DEPENDS ${VERSION_FILE})
+add_custom_target(${PROJECT_NAME}_VersionFile ALL DEPENDS ${VERSION_FILE})
 
 set(CPACK_PACKAGE_FILE_NAME "${_project_lower}-${PROJECT_VERSION}-${_sys}")
 set(CPACK_SOURCE_PACKAGE_FILE_NAME "${_project_lower}-${PROJECT_VERSION}")
--- a/cmake/install.cmake
+++ b/cmake/install.cmake
@@ -38,8 +38,21 @@ set(CPACK_RESOURCE_FILE_README "${CMAKE_CURRENT_SOURCE_DIR}/README")
 set(CPACK_PACKAGE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/package)
 string(TOLOWER ${CMAKE_SYSTEM_NAME} _sys)
 string(TOLOWER ${PROJECT_NAME} _project_lower)
-set(CPACK_PACKAGE_FILE_NAME "${_project_lower}-${git_version}-${_sys}")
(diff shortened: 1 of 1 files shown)


cburstedde:
Thanks! This is one more step to get the CMake tarball behave a bit more like the autotools one. Ideally, building with one system from a tarball created by the other should work.

Would we need analogous code in libsc?

tim-griesbach:
> Would we need analogous code in libsc?

Yes, if this PR is ready, I will create an analogous PR in libsc.

cburstedde:
I like the approach to replicate the autotools git-version-gen vs. .tarball-version behaviour as closely as possible. Is there any more to be done?

cburstedde:
> > Would we need analogous code in libsc?
> 
> Yes, if this PR is ready, I will create an analogous PR in libsc.

Cool, if it all works, let's take it there.

tim-griesbach:
> I like the approach to replicate the autotools git-version-gen vs. .tarball-version behaviour as closely as possible. Is there any more to be done?

The only remaining question for this PR is that the file `.tarball-version` is not created if one calls `make package_source` in a fresh repository, i.e. a repository that was never used to compile code. As I described [here](https://github.com/cburstedde/p4est/issues/325#issuecomment-2486285470), the usual way of expressing this dependency in CMake did not work for me.

For Autotools, `make dist` creates the `.tarball-version` file. On the other hand, `make dist` requires also a preceding call (`bootstrap`). 

Given that this issue seems to appear for t8code as well, I am interested in their perspective on this issue (cf. conversation in https://github.com/cburstedde/p4est/issues/325). If this does not lead to a solution, I could take a deeper dive and try to find a workaround.

cburstedde:
Thanks; should we merge at this point and fix the issue about calling some first build command later?

With p4est, calling bootstrap is only needed on a fresh repository checkout; on unpacking a tarball bootstrap is not required. How does CMake behave when executed from an unpacked tarball?

(We still produce different tarballs from automake and CMake. Hoping they'll get closer, eventually.)

tim-griesbach:
> Thanks; should we merge at this point and fix the issue about calling some first build command later?

Sounds good to me! I pushed one small change to ensure full symmetry to the analogous code in lisbc by using the CMake project name in the name of the target that generates `.tarball-version`. 

> With p4est, calling bootstrap is only needed on a fresh repository checkout; on unpacking a tarball bootstrap is not required. How does CMake behave when executed from an unpacked tarball?

In CMake the workflow from an unpacked tarball is the same as for a the git repository, i.e. calling the `cmake` command to configure and then build the code.

PR #327, opened 2024-11-19
Change default parameters in spheres2.c

hannesbrandt:
# More intuitive default parameters in spheres2.c

This PR changes some of the default  parameters in `example/spheres/spheres2.c` in order to create results that better fit the expectations for this example. In particular, we
 - set `maxlevel` above `minlevel`, so that the forest is adaptively refined,
 - increase `spherelems` to obtain higher resolution of the spheres and
 - increase `lfraction` to obtain multiple spheres.

Some of the spheres in this example overlap each other or lie partially outside of the forest, thereby demonstrating that the code is able to deal with these special cases as well.

### DIFF
--- a/example/spheres/spheres2.c
+++ b/example/spheres/spheres2.c
@@ -1073,16 +1073,16 @@ main (int argc, char **argv)
   /*** read command line parameters ***/
 
   opt = sc_options_new (argv[0]);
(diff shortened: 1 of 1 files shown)


cburstedde:
Thanks!

PR #328, opened 2024-11-25
Feature search doc

hannesbrandt:
# Doxygen documentation page for the search routines 

This PR adds `search.dox`, a documentation page for the search routines. It contains an overview and describes local and partition search in detail as well as the user-defined query points. To reference it as an example, we add documentation of `example/spheres` as well.


hannesbrandt:
That's a good point. I've added references to the 3D counterparts in the 2D examples.

doc/doxygen/example_particles.dox
```diff
@@ -130,7 +130,8 @@
 
 /** \example particles/particles2.c
  *
- * 2D version of a generic particle tracking demo.
+ * 2D version of a generic particle tracking demo (3D counterpart:
```
cburstedde, line 133: The first line shows up as a short title in the list on the web page. The information in parentheses is well meant, but maybe rather put into the text, not the first line, instead?

PR #333, opened 2024-12-17
seperat link command for Winsock

Davknapp:
# Title Use seperate link command for WINSOCK_LIBRARY

Following up on issue https://github.com/cburstedde/libsc/issues/206

Proposed changes: Use a seperate command instead of a generator expression. It prevents linker-errors with p4est if these expression fails to be properly generated.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,7 +46,11 @@ target_include_directories(p4est PUBLIC
   $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src>
   $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include>
   $<INSTALL_INTERFACE:include>)
-target_link_libraries(p4est PUBLIC SC::SC $<$<BOOL:${P4EST_HAVE_WINSOCK2_H}>:${WINSOCK_LIBRARIES}>)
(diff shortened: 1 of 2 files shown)


cburstedde:
Thanks again! I understand it's not originating from your code, but I'm wondering about an asymmetry with libsc: There we check for WIN32, here for HAVE_WINSOCK2_H. I would imagine this can be unified, with the goal to reduce the diff between the cmake environments of the two libraries as much as we possibly can.

lukasdreyer:
@cburstedde Are you willing to merge this without the change? We have some users running into problems with it. I can change `HAVE_WINSOCK2_H` into `WIN32` here, but I do not have any ideas of the technical implications.

CMakeLists.txt
```diff
@@ -48,7 +48,7 @@ target_include_directories(p4est PUBLIC
   $<INSTALL_INTERFACE:include>)
 target_link_libraries(p4est PUBLIC SC::SC )
 
-if (P4EST_HAVE_WINSOCK2_H)
+if ( WIN32)
```
cburstedde, line 51: Thanks, happy to merge! Is this supposed to be a space (in this case please remove) or an underscore?

Davknapp: The latest commit removes the space, looking forward to the merge :)

PR #337, opened 2025-02-07
Avoid CI fails due to busy file system

tim-griesbach:
# Avoid CI fails due to busy file system

This PR provides analogous changes as implemented in libsc in the PR https://github.com/cburstedde/libsc/pull/214. This means that this PR adds an option in Autoconf and in CMake to disable the file I/O checks, which cause frequent CI failures due to a busy file system on the CI runners. This option is used in the three CI workflows.

In addition, this PR adds `test_loadsave3` as target in CMake since it was missing and this PR fixes a typo in `CMakeLists.txt` that caused the CMake build to fail on Windows.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,11 @@ on:
       - ".github/workflows/ci_valgrind.yml"
   pull_request:
 
+env:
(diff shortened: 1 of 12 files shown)


cburstedde:
Thanks!

PR #339, opened 2025-02-17
Unify Winsock2 checks in p4est and libsc

hannesbrandt:
# Unify Winsock2 checks in p4est and libsc

Following up on the asymmetry of Winsock inclusion between p4est and libsc, which was mentioned in #333. 

This PR replaces P4EST_HAVE_WINSOCK2_H with checks for WIN32, similar to what was done for libsc in https://github.com/cburstedde/libsc/pull/150. Additionally, we now define WIN32_LEAN_AND_MEAN, to avoid Winsock.h from being included when Winsock2.h is included. Thanks to @scivision for his offline support.

We would appreciate feedback from @sloede and @lukasdreyer, if these changes somehow affect Windows portability on their end.


cburstedde:
Thanks!

PR #340, opened 2025-02-19
Add generic geometry transformation

cburstedde:
# Generic geometry transformation

## Proposed changes

So far, calling a geometry map was left to the user and required reading quite some documentation.
Here we introduce the function p4est_geometry_transform_coordinates that does the job.
It can be called given any valid geometry structure to map quadrant reference coordinates into space.

Along the way, we add minor documentation fixes.

Any comments @pkestene ?

### DIFF
--- a/example/simple/simple2.c
+++ b/example/simple/simple2.c
@@ -239,36 +239,25 @@ static int
 refine_icosahedron_fn (p4est_t *p4est, p4est_topidx_t which_tree,
                        p4est_quadrant_t *quadrant)
 {
-
   p4est_geometry_t   *geom = (p4est_geometry_t *) p4est->user_pointer;
 
   /* logical coordinates */
-  double              xyz[3] = { 0, 0, 0 };
+  p4est_qcoord_t coords[2];
 
   /* physical coordinates */
-  double              XYZ[3] = { 0, 0, 0 };
+  double              XYZ[3];
 
-  double              h2 =
-    0.5 * P4EST_QUADRANT_LEN (quadrant->level) / P4EST_ROOT_LEN;
-  const double        intsize = 1.0 / P4EST_ROOT_LEN;
+  /* from logical coordinates to physical coordinates (Cartesian) */
+  p4est_quadrant_volume_coordinates (quadrant, coords);
+  p4est_geometry_transform_coordinates (geom, which_tree, coords, XYZ);
 
-  /*
-   * get coordinates at cell center
-   */
-  xyz[0] = intsize * quadrant->x + h2;
-  xyz[1] = intsize * quadrant->y + h2;
-#ifdef P4_TO_P8
-  xyz[2] = intsize * quadrant->z + h2;
-#endif
-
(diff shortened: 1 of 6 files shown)


cburstedde:
I'm wondering about refine_pillow_fn in simple3.c. I tried to update the refinement criterion equivalently. Yet I'm wondering about the choice of the two reference points, where the second gets a half-shift in x. Would it make sense to change the second point for some more symmetry, or is this exactly as intended?

cburstedde:
About the prototype of the coordinates functions: upstream it is coords[]; more consistent with the rest of the code would be coords[2] and coords[3] in the .h files, and [P4EST_DIM] in the .c files, respectively. Would @tisaac see any risk in interfacing to outside code in making this change?

cburstedde:
> About the prototype of the coordinates functions: upstream it is coords[]; more consistent with the rest of the code would be coords[2] and coords[3] in the .h files, and [P4EST_DIM] in the .c files, respectively. Would @tisaac see any risk in interfacing to outside code in making this change?

Indeed, this would be a bigger update and discussion needed. Let's not mix it with this PR.

pkestene:
> I'm wondering about refine_pillow_fn in simple3.c. I tried to update the refinement criterion equivalently. Yet I'm wondering about the choice of the two reference points, where the second gets a half-shift in x. Would it make sense to change the second point for some more symmetry, or is this exactly as intended?

Hi Carsten, I'm ok with the new API about coordinates transform, cool.

Regarding `refine_pillow_fn`, the intent was just to use radius as the refine criterium.  By looking at `p8est_geometry_pillow_X`, one can notice that actually it is `rst[2]` that maps the radial location between `R1`and `R2`. So maybe, the shift should better be along `quadrant->z` instead of `quadrant->x`. Thanks for noticing that.
It just realized that, I was a bit lazy here. A more correct design could be to take the first point at the center of -Z face, and the second point at the center of the +Z face.

Visually, it won't change the mesh much, but it makes more sense. If you agree, you can change it accordingly.

cburstedde:
> > I'm wondering about refine_pillow_fn in simple3.c. I tried to update the refinement criterion equivalently. Yet I'm wondering about the choice of the two reference points, where the second gets a half-shift in x. Would it make sense to change the second point for some more symmetry, or is this exactly as intended?
> 
> Hi Carsten, I'm ok with the new API about coordinates transform, cool.
> 
> Regarding `refine_pillow_fn`, the intent was just to use radius as the refine criterium.  By looking at `p8est_geometry_pillow_X`, one can notice that actually it is `rst[2]` that maps the radial location between `R1`and `R2`. So maybe, the shift should better be along `quadrant->z` instead of `quadrant->x`. Thanks for noticing that.
> It just realized that, I was a bit lazy here. A more correct design could be to take the first point at the center of -Z face, and the second point at the center of the +Z face.

Thanks @pkestene, this is exactly the information I was looking for!

> Visually, it won't change the mesh much, but it makes more sense. If you agree, you can change it accordingly. 

I've updated the PR accordingly.  It looks beautiful, thanks!

Issue #295, opened 2024-03-14
Place connectivity in shared memory


cburstedde:
Currently the p4est connectivity structure is replicated on all processes.
If the connectivity is large, this wastes memory on multicore nodes.
Thus create an alternative allocation function that uses MPI-3 shared memory.

cburstedde:
Ok this took me a while to begin. It required just 7 hours of coding. :)

Feedback is very welcome.

cburstedde:
Closing this unless there's anything left to be desired wrt. this feature.

PR #344, opened 2025-03-07
Feature conn shared

cburstedde:
# Share a connectivity using MPI-3

Following up on issue #295 

## Proposed changes

We have added p4est_connectivity_new_shared to place a connectivity that exists on one rank of a shared memory node in that node's MPI-3 shared memory. This saves much memory on nodes with many cores.

To get the connectivity on each node's first rank in the first place, users may call p4est_connectivity_bcast with a communicator of all such ranks across the world.

Any comments are much appreciated. @ackirby88


ackirby88:
Thank you, @cburstedde! 
I haven't fully tested this yet, but I did notice that p4est_load/p4est_load_ext/p4est_load_mpi may need to be extended also.

cburstedde:
> Thank you, @cburstedde! I haven't fully tested this yet, but I did notice that p4est_load/p4est_load_ext/p4est_load_mpi may need to be extended also.

Thanks for the feedback! I've pushed a use case to test/test_valid2 including the proposed high-level API function p4est_connectivity_mission. Comments before merging the PR?

Feedback would be very welcome wrt. identical behavior under autotools and CMake as well as correct behavior (a) without MPI, (b) with MPI but without MPISHARED and (c) with MPISHARED.

Good point on the load functions. I'd be leaning towards adding separate functions p4est_load_share and p4est_source_share in a separate PR, will investigate.

ackirby88:
> Thanks for the feedback! I've pushed a use case to test/test_valid2 including the proposed high-level API function p4est_connectivity_mission. Comments before merging the PR?

I was just coding this up on my end, so this is perfect for making it broadly usable by the community. 

> Feedback would be very welcome wrt. identical behavior under autotools and CMake as well as correct behavior (a) without MPI, (b) with MPI but without MPISHARED and (c) with MPISHARED. 
 
I have some partial issues to work out on my end for unstructured meshes (e.g., broadcasting geometries & periodic vertices), which I hope to get to towards the end of this week.

cburstedde:
This is the release candidate; happy for any feedback.

cburstedde:
This branch is current with the recent master update. I have not tested this with a true multi-node machine: feedback in that regard would be awesome.

cburstedde:
> > Thanks for the feedback! I've pushed a use case to test/test_valid2 including the proposed high-level API function p4est_connectivity_mission. Comments before merging the PR?
> 
> I was just coding this up on my end, so this is perfect for making it broadly usable by the community. 

By our design the MPI-3 window is spread as evenly as possible over all
ranks of a node, aligned to the requested data structure unit size.
Correspondingly, we call MPI_WIN_LOCK_ALL after writing the data.

The sc/sc_shmem.c approach allocates all memory on node rank zero and uses
MPI_WIN_LOCK (0).  Which is preferred?  Any evidence/experience/opinion?

cburstedde:
Inclined to merge right now for release and to continue afterwards.
Especially since the forest load functions should be updated.

Feel free to reopen/comment anytime.

PR #345, opened 2025-03-09
Feature distro

cburstedde:
# Meet requirements of distributions

Some Linux distributions require an arch-independent p4est_config.h file.

## Proposed changes

Remove compiler and flag definitions from the configuration header, and do not rely on configure-time endian checks anymore. Any comments, especially wrt. pkgconfig? @tamiko

@scottaiton I might need some help checking that I did not break the VTK output.


tamiko (approved):
This looks good.

tamiko:
@cburstedde The pkgconf part looks reasonable - I'll check what what happens with this configuration change on debian and gentoo today or tomorrow and let you know if something needs fixing.

cburstedde:
> @cburstedde The pkgconf part looks reasonable - I'll check what what happens with this configuration change on debian and gentoo today or tomorrow and let you know if something needs fixing.

Great, thanks for your help!

cburstedde:
> > @cburstedde The pkgconf part looks reasonable - I'll check what what happens with this configuration change on debian and gentoo today or tomorrow and let you know if something needs fixing.

We've updated libsc to latest develop and pushed it onto this candidate.

cburstedde:
What I noticed is that CMake puts .tarball_version into the source directory.
That file should rather go into the build directory. Is that fixable short notice?

cburstedde:
I'm merging now, let's keep an eye on .tarball_version. Thanks for checking!

PR #347, opened 2025-03-28
Update README: Install instructions

DanielDoehring:
# Update README: Install instructions

Following up on issue # . -

Proposed changes:

I would add the `make`, `make install` commands for completeness (and especially for people less familiar with the GNU build process, e.g. for folks coming from Julia.)

### DIFF
--- a/README
+++ b/README
@@ -145,7 +145,7 @@ tree.  For development or for testing multiple configure options, create an
 empty directory for each configuration and invoke configure with a relative
 path.
 Note that the `--disable-shared` option suppresses the build of `.so` shared object lib files.
-In case these are also required the option needs to be omitted or changed to `--enable-shared`.
+In case these are also required this option needs to be omitted or changed to `--enable-shared`.
 
 The `configure` call is then followed by the usual `make` and `make install` commands, 
 potentially in `sudo` mode depending on the desired installation directory.
--- a/README
+++ b/README
@@ -148,7 +148,7 @@ Note that the `--disable-shared` option suppresses the build of `.so` shared obj
 In case these are also required this option needs to be omitted or changed to `--enable-shared`.
 
 The `configure` call is then followed by the usual `make` and `make install` commands, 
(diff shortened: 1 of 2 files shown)


README
```diff
@@ -144,6 +144,11 @@ installation, `./configure` can be called in the root directory of the unpacked
 tree.  For development or for testing multiple configure options, create an
 empty directory for each configuration and invoke configure with a relative
 path.
+Note that the `disable-shared` suppresses the build of `.so` shared object lib files.
```
DanielDoehring, line 147:
```suggestion
Note that the `--disable-shared` option suppresses the build of `.so` shared object lib files.
```

README
```diff
@@ -144,6 +144,11 @@ installation, `./configure` can be called in the root directory of the unpacked
 tree.  For development or for testing multiple configure options, create an
 empty directory for each configuration and invoke configure with a relative
 path.
+Note that the `--disable-shared` option suppresses the build of `.so` shared object lib files.
+In case these are also required the option needs to be changed to `--enable-shared`.
```
DanielDoehring, line 148:
```suggestion
In case these are also required the option needs to be omitted or changed to `--enable-shared`.
```

README
```diff
@@ -144,6 +144,11 @@ installation, `./configure` can be called in the root directory of the unpacked
 tree.  For development or for testing multiple configure options, create an
 empty directory for each configuration and invoke configure with a relative
 path.
+Note that the `--disable-shared` option suppresses the build of `.so` shared object lib files.
+In case these are also required the option needs to be omitted or changed to `--enable-shared`.
```
DanielDoehring, line 148:
```suggestion
In case these are also required this option needs to be omitted or changed to `--enable-shared`.
```

cburstedde:
Thanks for your contributions!

Even though this is just doc and not code, would you like to add a file doc/author_<lastname>.txt releasing under BSD?

PR #348, opened 2025-04-09
Unify example executable names

hannesbrandt:
# Unify example executable names

This PR changes the naming scheme of the executables in example/CMakeLists.txt to match the `p4est_` naming scheme used in Autotools.
Additionally, it adds the userdata example to example/CMakeLists.txt.
Furthermore, it adds some missing files from the example folders to the autotools distribution.

### DIFF
--- a/example/points/Makefile.am
+++ b/example/points/Makefile.am
@@ -7,14 +7,14 @@ if P4EST_ENABLE_BUILD_2D
 bin_PROGRAMS += example/points/p4est_points
 example_points_p4est_points_SOURCES = example/points/points2.c
 
-bin_PROGRAMS += example/points/p4est_points_generate
-example_points_p4est_points_generate_SOURCES = example/points/generate_points2.c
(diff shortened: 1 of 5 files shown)


example/points/Makefile.am
```diff
@@ -7,14 +7,14 @@ if P4EST_ENABLE_BUILD_2D
 bin_PROGRAMS += example/points/p4est_points
 example_points_p4est_points_SOURCES = example/points/points2.c
 
-bin_PROGRAMS += example/points/p4est_points_generate
```
cburstedde, line 10: In this file I'd rather change the names on the CMake side, not the automake side.

hannesbrandt: Okay, I have changed the name back to points_generate

cburstedde: Thanks; I see that we may want to rename just the .c files generate_points eventually, here we are good.

PR #353, opened 2025-06-06
CI:Windows:MSYS2: use UCRT packages

scivision:
# CI use MSYS2 UCRT environment prefix

Proposed changes:
UCRT is the recommend environment prefix of the MSYS2 project and what Microsoft promotes as well. UCRT is what MSYS2 end-users would be using. This makes the CI tests best represent what end users experience.


cburstedde:
Thanks! Would it make sense to do this for libsc analogously?

PR #355, opened 2025-07-23
Do not mark functions in header files as 'static'.

hannesbrandt:
# Do not mark functions in header files as 'static'.

Analogous to https://github.com/cburstedde/libsc/pull/226, this PR removes the 'static' keyword from inline functions in p4est's .h files.


cburstedde:
> I've removed the ***@***.***@` for the inline functions and updated their documentation. There are also some functions affected by this PR that did not have the ***@***.***@` in their documentation and are currently undocumented, which leads to a doxygen warning. Should I add documentation for these functions as well?

If you can that would be awesome.

PR #364, opened 2026-05-22
Fix unused variable warnings

hannesbrandt:
# Fix set-but-unused-variable warnings

Proposed changes:
This PR addresses a recent issue in the CI (https://github.com/cburstedde/p4est/actions/runs/26040411092/job/76567611984). There were some set-but-unused-variable warnings for variables that are only used in assertions. This PR removes the warnings by enclosing the variables in an #ifdef P4EST_ENABLE_DEBUG. Also, it updates the submodule to include similar fixes in sc (see https://github.com/cburstedde/libsc/pull/240).


cburstedde:
Thanks again!

Issue #365, opened 2026-06-18
p4est_wrap_adapt returns false even though mesh has changed


scottaiton:
@donnacalhoun and I have have ran into an issue in ForestClaw where `p4est_wrap_adapt` is returning false even though the mesh has changed. This causes ForestClaw to not update for the new mesh, and eventually causes ForestClaw to segfault.

I think the problem essentially is that that `p4est_wrap_adapt` uses a checksum, and there is a checksum collision between the original mesh and the new mesh.

This problem currently is occurring in the `clawpack/shallow/2d/bump` and the `clawpack/advection/2d/hemisphere` examples currently in `forestclaw/develop`.

donnaaboise:
A third example is in `forestclaw/applications/clawpack/shallow/2d/radialdam`.   Here is the *.ini file that generates the error : 

```
[user]
     # 0  : Cartesian grid
     # 1  : Five patch square
     # 2  : Pillow disk
     # 3  : Pillow disk mapping of the five patch square
     example = 1

     alpha = 0.4     # For five-patch disk

     x0 = 0
     y0 = 0
     r0 = 0.5
     hin = 2.0
     hout = 1.0      

     claw-version = 4

[clawpatch]
     # Grid dimensions
     mx = 16               # mx_leaf
     my = 16               # my_leaf
     mbc = 2              # Number of ghost cells
     maux = 16

     # Clawpatch tagging criteria
     # value       : value exceeds threshold
     # minmax      : qmax-qmin exceeds threshold
     # difference  : difference (e.g. dqx = q(i+1,j)-q(i-1,j)) exceeds threshold
     # gradient    : gradient exceeds threshold
     # user        : User defined criteria
     refinement-criteria = gradient

     meqn = 4         # Number of equations

[Options]

# Regridding information
     minlevel = 0               # Minimum level
     maxlevel = 3               # Maximum levels of refinement

     regrid_interval = 1        # Regrid every 'regrid_interval' time steps.
     refine_threshold = 1.0
     coarsen_threshold = 0.5  # 0.41111
     smooth-refine = T
     smooth-level = 6

# Time stepping
     tfinal = 1.5
     initial_dt = 0.015625            # Initial time step for 'minlevel'

     use_fixed_dt = F            # Take a fixed time step
     max_cfl = 1.0
     desired_cfl = 0.9

     outstyle = 1                # OutputStyle = 1,2,3 (see below, e.g. [OutputStyle_1]
[...]

PR #366, opened 2026-06-18
add example for p4est_wrap_adapt checksum collision

scottaiton:
Following up on issue #365 .

Here is an example demonstrating the issue. I isolated these meshes from a ForestClaw example that was failing.


cburstedde:
Thanks; if you'd like to keep this on top of the latest develop branch, I'd just be wondering whether the CI can be made to run.

Issue #319, opened 2024-10-14
Drop 3d connectivity broken ?


pkestene:
**Description**

When running `./p8est_simple drop 5` (from example/simple sub-directory), one obtains the following log

```
...
[p4est 0]  Balance tree 3 A 8912
[p4est 0]  Into balance tree 4 with 1065
[p4est 0]  Hash size 255 avg 0.0314 std 0.196 checks 0 0
[p4est 0]  Hash size 255 avg 0.141 std 0.37 checks 0 0
[p4est 0]  Tree 4 inlist 0 outlist 686 ancestor 19 insert 161
[p4est 0]  Balance tree 4 A 1226
[libsc 0] Abort: Assertion 'p4est_quadrant_is_ancestor (tq, s) || (p4est_quadrant_is_equal (tq, s) && tq->level == P4EST_QMAXLEVEL)'
[libsc 0] Abort: ../src/p4est_algorithms.c:1062
[libsc 0] Abort: Obtained 11 stack frames
...
```

**To Reproduce**

p4est configured with `./configure --enable-mpi --enable-debug CFLAGS=-O0 CC=mpicxx`

**Additional information**

I checked current HEAD (develop) and v2.8.6 and v2.8.5 with the same results.

pkestene:
Is it possible (somehow) that, after the refined step, the octree is "too" off-balanced that `p4est_balance` is unable to fix it ?

cburstedde:
> Is it possible (somehow) that, after the refined step, the octree is "too" off-balanced that `p4est_balance` is unable to fix it ?

Good find. The drop connectivity is indeed unusual. We had previously determined that every connectivity always needs full edge and corner information to support balance, even if balance is not extended across edges and/or faces. So we'd have to determine first whether the drop connectivity is complete. Any volunteers @pkestene @tim-griesbach ?

cburstedde:
> Is it possible (somehow) that, after the refined step, the octree is "too" off-balanced that `p4est_balance` is unable to fix it ?

Thanks for the report!  Generally, balance is designed such that absolutely
any mesh is legal as input.  We'll have to double-check the drop
connectivity, and if the error persists look deeper into balance.  Which is
on our list already, but we tend to get distracted since it usually works.

tim-griesbach:
[...]

Issue #101, opened 2021-03-12
Tighted owned neighborhood check


cburstedde:
**Description**

The branch `test-neighborhood-owned` contains work in progress to improve testing the 3x3 quadrant neighborhood. We're on the way removing unneeded information when edge and/or corner connections are not requested. This would need to be finalized, the checks reduced to the minimal set of cases, and tested.

**Proposed solution**

Talk to me about where to begin and how to proceed with this branch.

cburstedde:
The tip of my code to make neighborhood checks conection-type-specific is ae4db19720f01d111aa0b229d936724836003c9b.

This may need double-checking and further integration.

cburstedde:
In view of the principle formulated in #112, it remains to check whether this branch is an actual improvement or a red herring.

PR #368, opened 2026-07-03
Fix balance

hannesbrandt:
Following up on issue https://github.com/cburstedde/p4est/issues/238 and https://github.com/cburstedde/p4est/issues/319 . As discussed in https://github.com/cburstedde/p4est/issues/238, balance (and ghost) use face-only  tree connections when creating the `borders` array at some places independent of the `btype`. This leads to unbalanced results of `p4est_balance`, e.g. for trees that are connected only by edge/corner.
In this PR `p4est_comm_neighborhood_owned` is changed such that all quadrants touching the tree boundary are added to the borders array even if there is no neighboring tree and the function is renamed to `p4est_comm_is_boundary`.

The face-only tree connections are also used to identify isolated trees. Therefore, in a connectivity of two trees connected by a single edge or corner both would erroneously be identified as isolated and skipped. To address this skipping of isolated trees was removed. Also, this PR adds a new connectivity edge_corner, which is used to test balance in the mentioned corner cases. It contains three trees with one edge and one corner connection between them.

The described fixes come with a performance regression ranging from 0%-20% according to our tests. This is most likely due to the borders array, which increases in size by the less strict version of `p4est_comm_neighborhood_owned` . The second fix (no longer skipping isolated trees) should only have an impact on rare connectivites with a high percentage of isolated trees.

Proposed changes:
- Fix corner cases in balance and ghost.
- Rename `p4est_comm_neighborhood_owned` to `p4est_comm_is_boundary` .
- Accept all tree boundary quadrants in `p4est_comm_is_boundary` and switch return values to better fit the new name.
- Add a new 3d connectivity edge_corner of trees connected only across edges and corners.
- Add unit tests for balancing purely edge- or corner-connected trees.

@tisaac


hannesbrandt:
> The described fixes come with a performance regression ranging from 0%-20% according to our tests. This is most likely due to the borders array, which increases in size by the less strict version of `p4est_comm_neighborhood_owned` . The second fix (no longer skipping isolated trees) should only have an impact on rare connectivites with a high percentage of isolated trees.

Of course, we could avoid this regression by checking for face, edge and corner tree connections depending on the btype at the relevant places (cf. https://github.com/cburstedde/p4est/issues/101).

cburstedde:
> > The described fixes come with a performance regression ranging from 0%-20% according to our tests. This is most likely due to the borders array, which increases in size by the less strict version of `p4est_comm_neighborhood_owned` . The second fix (no longer skipping isolated trees) should only have an impact on rare connectivites with a high percentage of isolated trees.
> 
> Of course, we could avoid this regression by checking for face, edge and corner tree connections depending on the btype at the relevant places (cf. https://github.com/cburstedde/p4est/issues/101).

Thanks, @hbrandt and @tim-griesbach.  I'd prefer merging this one first, to
harden balance against rare/evil connectivities, and to add the optimization
(checking tree contacts, this time for all codimensions) at a later point.

cburstedde:
Thanks again!

